PR #5, opened 2015-03-25
H5MD writer

MrTheodor:


examples/polymer_melt/polymer_melt.py
```diff
@@ -1,5 +1,5 @@
-#!/usr/bin/env python
```
junghans: Why did you actually modify this file?

src/CMakeLists.txt
```diff
@@ -43,8 +46,23 @@ endif (WITH_RC_FILES)
...
-add_library(_espressopp ${ESPRESSO_SOURCES})
-target_link_libraries(_espressopp ${Boost_LIBRARIES} ${PYTHON_LIBRARIES} ${MPI_LIBRARIES} ${FFTW3_LIBRARIES} ${VAMPIRTRACE_LIBRARIES})
+if (EXTERNAL_BOOST)
+  set (BOOST ${Boost_LIBRARIES} )
+else(EXTERNAL_BOOST)
+  set (BOOST espresso_boost)
+endif(EXTERNAL_BOOST)
+
+if(WITH_H5MD)
+  find_package(HDF5 REQUIRED COMPONENTS "C")
```
junghans: I think, you need the `CXX` component.

src/Version.cpp
```diff
@@ -34,28 +35,36 @@
...
-	  date          = __DATE__;
-	  time          = __TIME__;
+    name_          = "ESPResSo++";
+    major_         = MAJORVERSION;
+    minor_         = MINORVERSION;
+    patchlevel_    = PATCHLEVEL;
+    gitrevision_    = gitversion;
+    boostversion_  = BOOST_LIB_VERSION;
+    date_          = __DATE__;
+    time_          = __TIME__;
```
junghans: Why?

examples/polymer_melt/polymer_melt.py
MrTheodor: No reason, some leftovers, reverting this file.

src/io/DumpH5MD.cpp
```diff
@@ -0,0 +1,262 @@
+/*
+  Copyright (c) 2015
+      Jakub Krajniak (jkrajniak at gmail.com)
+  Copyright (C) 2012,2013
```
junghans: I think these lines can be dropped as the reader was added in 2015 and didn't existed back in 2008-2013.

src/io/DumpH5MD.hpp
```diff
@@ -0,0 +1,105 @@
+/*
+  Copyright (C) 2012,2013
```
junghans: Copyright wrong.

src/io/__init__.py
```diff
@@ -2,21 +2,21 @@
 #      Max Planck Institute for Polymer Research
 #  Copyright (C) 2008,2009,2010,2011
 #      Max-Planck-Institute for Polymer Research & Fraunhofer SCAI
-#  
+#
```
junghans: Why?

src/io/ch5md.cpp
```diff
@@ -0,0 +1,431 @@
+/* This file is part of ch5md
+ * Copyright (C) 2013-2014 Pierre de Buyl
+ * All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the BSD license. See the LICENSE file for details.
```
junghans: If this part is licensed under BSD (and not GPL-3 like the rest of Espresso) it should be noted in the `README` and `COPYING`. See https://www.softwarefreedom.org/resources/2007/gpl-non-gpl-collaboration.html

junghans:
I think `h5md_writer/polymer_melt.lammps` should not be copying, but made a symlink. Adding a test in the testsuite and enabling it in `.travis.yml` would be nice, too.

src/Version.cpp
MrTheodor: 1. clean up, there was a mix of tabs and spaces.
2. the variable name 'name_' allows me to have ver.name() in the code, of course that's a matter of taste, nothing ore. (following: https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Variable_Names)

junghans: I don't really have a strong opinion, but I don't like changing stuff just for the sake of indentation and style guides. @stuehn , up to you.

Maybe we should run `uncrustify` and google style checker on the whole code in a separate change.

MrTheodor: Sure, committing only trailing space is not a good idea. But the rule 'touched-owned' really improve the style of code.
I think that running one-time such tool without any clear style code rule will not make any difference in long term developing. That's rather a politics and project organization. E.g. allow obsolete code to be on the `master` branch and so on.

.travis.yml
```diff
@@ -6,14 +6,26 @@ before_install:
  - sudo apt-get update
  - sudo apt-get install cmake libfftw3-dev libopenmpi-dev openmpi-bin
  - if [[ $EXTERNAL = ON ]]; then sudo apt-get install libboost1.55-all-dev python-mpi4py; fi
+ - if [[ $H5MD = ON ]]; then
+        wget http://www.hdfgroup.org/ftp/HDF5/current/src/hdf5-1.8.14.tar
```
junghans: I think you want to connect all these lines with `&&`.

.travis.yml
```diff
@@ -6,14 +6,26 @@ before_install:
...
+        tar -xzvf hdf5-1.8.14.tar
+        cd hdf5-1.8,14
```
junghans: I think `1.8.14` instead of `1.8,14`.

pdebuyl:
Hi,

The files ch5md.hpp and ch5md.cpp cannot carry the GPL license header. They can be included in espressopp indeed provided that the licensing information is preserved. Also, they can have the h and c extensions as the code is declared as extern c. The copyright can be updated as Jakub made changes.

    #ifdef __cplusplus
    extern "C" {
    #endif

.travis.yml
```diff
@@ -6,14 +6,26 @@ before_install:
...
+        ./configure --enable-parallel --enable-shared --prefix=/usr
+        make && make install
+        pip install h5py
+    fi
+
+   fi
```
pdebuyl: What `if` is this `fi` closing?

junghans:
I checked that again, including BSD parts in GPL-3 codes is fine. And BSD doesn't require a `NOTICE` file, so everything is good. However, I would still mention the code contribution in `README.md`.

MrTheodor:
Retracted as Pierre's approach is simpler.

PR #16, opened 2015-10-07
Revert to a working version of LB. 

niktre:
Revert to a working version of LB. Added a parameter for profiling from the python script.

### DIFF
--- a/examples/lattice_boltzmann/lb_performance.py
+++ b/examples/lattice_boltzmann/lb_performance.py
@@ -1,0 +1,43 @@
+# DEMONSTRATION OF THE LATTICE-BOLTZMANN SIMULATION
+#
+s=20
+import espressopp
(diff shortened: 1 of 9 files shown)


examples/lattice_boltzmann/lb_performance.py
```diff
@@ -0,0 +1,43 @@
...
+lb.lbTemp = 1.
+lb.nSteps = 1
+## add some profiling statistics for the run
+for k in range (3):
+#	lb.readCouplForces()
+	integrator.run(50000)
+	s = str(integrator.step)
+	mdoutput = 'dump.' + s + '.xyz'
+	espressopp.tools.writexyz(mdoutput, system)
+	lb.saveCouplForces()
```
junghans: Could you add a test case in `/testsuite` directory, too?

PR #17, opened 2015-10-08
cmake: make BUILD_SHARED_LIBS=OFF work again

junghans:


junghans:
We need to do something here, currently `cmake` just throws an syntax error message if `BUILD_SHARED_LIBS=OFF` and that is bad, because it leaves the user alone. I think we can either throw in error message (like "static modules not supported") in that case and hence disallow static linking or include a patch like the one above and leave to the user to figure out how get python to load static module (which was the problem at the tutorial yesterday).

govarguz:
True. We can go for option 1 as a first approach, i.e. throw an error message warning the user about the static modules related to the package.

PR #18, opened 2015-10-12
Enabling  Multiple Communicators for PMI

govarguz:
Added check for PMI multiple communicators in the python modules


junghans:
To be honest, wtf, why do you have 15 times the same commit message?

govarguz:
It was in the queue systematically added. Besides one more ss about to come with a very similar message.However, I will use rather the new methodology you have mentioned yesterday evening.

PR #19, opened 2015-10-14
Further enabling multiple communicator

govarguz:
Further enabling multiple communicator. All within one commit!


MrTheodor:
Sorry to bother, but is this line not implemented in pmi.workerIsActive already ?

govarguz:
good point! the added line do not apply to all files handling communicators

niktre:
Why did you kill all David's documentation lines and reverted to the original pre-David style with strange 'The (local) fixed pair list.'???

junghans:
Copyright years need update, too!

PR #22, opened 2015-10-26
New CoulombTruncated class

acfogarty:
Moved old CoulombTruncated to CoulombTruncatedUniqueCharge and created a new more flexible CoulombTruncated class which allows different charges on different particles.


MrTheodor:
It looks nice!

PR #27, opened 2015-11-11
added new Dihedral Harmonic potential functional 2 and corrected the …

govarguz:
added new Dihedral Harmonic potential functional 2 and corrected the sign of the existing Dihedral Harmonic Ncos potential. Plus some cosmetic changes to topology helper and gromacs.py

### DIFF
--- a/src/interaction/DihedralHarmonic.hpp
+++ b/src/interaction/DihedralHarmonic.hpp
@@ -138,7 +138,7 @@ namespace espressopp {
         /// The part of the formula. 1/sin(phi) * d/dphi U(phi)
         /// where the \f$U(\phi) = 0.5 K (\phi_{ijkn} - \phi_0)^2]\f$
         ///
-        /// The derivative of \f$U(phi)\f$ is \f$K*n*sin(\phi_0 - n*\phi)\f$
+        /// The derivative of \f$U(phi)\f$ is \f$K*sin(\phi_0 - \phi)\f$
         ///
         real coef1 = (1.0/sin(_phi)) * (K * (_phi - phi0));
 
@@ -189,6 +189,7 @@ namespace espressopp {
        * @return The value of the force
        */
       real _computeForceRaw(real phi) const {
+	real sin_phi = sin(phi);
 	real coef1 = (1.0/sin(phi)) * K * (phi - phi0);
         return -1.0 * coef1;
       }
--- a/src/interaction/DihedralHarmonic.hpp
+++ b/src/interaction/DihedralHarmonic.hpp
@@ -47,7 +47,6 @@ namespace espressopp {
     private:
       real K;
       real phi0;
-      //real cos_phi0;
 
     public:
       static void registerPython();
@@ -138,7 +137,7 @@ namespace espressopp {
         /// The part of the formula. 1/sin(phi) * d/dphi U(phi)
         /// where the \f$U(\phi) = 0.5 K (\phi_{ijkn} - \phi_0)^2]\f$
         ///
-        /// The derivative of \f$U(phi)\f$ is \f$K*sin(\phi_0 - \phi)\f$
+        /// The derivative of \f$U(phi)\f$ is \f$K*(\phi_0 - \phi)\f$
         ///
         real coef1 = (1.0/sin(_phi)) * (K * (_phi - phi0));
 
@@ -190,6 +189,10 @@ namespace espressopp {
        */
       real _computeForceRaw(real phi) const {
 	real sin_phi = sin(phi);
+        if (fabs(sin_phi) < 1e-9) {
+          if (sin_phi>0.0) sin_phi = 1e-9;
+	  else sin_phi = -1e-9;  	
+	}
 	real coef1 = (1.0/sin(phi)) * K * (phi - phi0);
         return -1.0 * coef1;
       }
--- a/src/interaction/DihedralHarmonic.hpp
+++ b/src/interaction/DihedralHarmonic.hpp
@@ -1,7 +1,5 @@
 /*
-  Copyright (C) 2014
-      Jakub Krajniak
-  Copyright (C) 2012,2013
+  Copyright (C) 2012,2013,2015 (H)
       Max Planck Institute for Polymer Research
   Copyright (C) 2008,2009,2010,2011
       Max-Planck-Institute for Polymer Research & Fraunhofer SCAI
@@ -23,8 +21,8 @@
 */
 
 // ESPP_CLASS
-#ifndef _INTERACTION_DIHEDRALHARMONICNCOS_HPP
-#define _INTERACTION_DIHEDRALHARMONICNCOS_HPP
+#ifndef _INTERACTION_DIHEDRALHARMONIC_HPP
+#define _INTERACTION_DIHEDRALHARMONIC_HPP
 
 #include "DihedralPotential.hpp"
 #include <cmath>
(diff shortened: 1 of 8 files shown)


src/interaction/DihedralHarmonic.hpp
```diff
@@ -0,0 +1,201 @@
...
+        force1 = retF[0];
+        force2 = retF[1];
+        force3 = retF[2];
+        force4 = retF[3];
+      }
+
+      /*
+       * Compute raw force for the tabulated potential, depend only on the phi value
+       * \f$F(\phi) = -grad V = -grad [ 0.5 (\phi_{ijkn} - \phi_0)^2] ]  \f$
+       * where n is a multiplicity and \f$phi_0\f$ is the reference angle
```
acfogarty: We can get rid of the reference to multiplicity here.

src/interaction/DihedralHarmonic.hpp
```diff
@@ -0,0 +1,201 @@
...
+       *
+       * @param[in] phi Degree in radians
+       *
+       * @return The value of the force
+       */
+      real _computeForceRaw(real phi) const {
+	real coef1 = K * (phi - phi0) + 0.5 * K * phi0 * phi0;
```
acfogarty: Are you sure there's a 0.5*K*phi0*phi0 term in grad [ 0.5 (\phi_{ijkn} - \phi_0)^2] ] ? Or maybe I'm misunderstanding something.

src/interaction/DihedralHarmonic.hpp
```diff
@@ -0,0 +1,201 @@
...
+        return -1.0 * coef1 ;
```
acfogarty: Separate issue: what are these _computeForceRaw(real phi) functions in the different analytical potentials actually used for? And what are they supposed to return? Here and in HarmonicNCosDihedral (after your bugfix in this commit) they return -grad V. But in other classes e.g. DihedralHarmonicCos it seems to return -gradV*1/sin(phi). This corresponds to the scalar prefactor of the force (e.g. see how it's written in the DL_POLY manual). The class Harmonic for bonds also returns -gradV*1/sin(theta), not just -gradV.

I presume the output of these _computeForceRaw(real phi) functions is supposed to correspond to the variable a in TabulatedDihedral, whatever that is...

Maybe Jakub has some idea about this, I'll also tag him.

src/interaction/DihedralHarmonic.py
```diff
@@ -0,0 +1,123 @@
...
+*********************************************************
+
+The dihedral harmonic potential
+
+.. math::
+
+   U(\phi_{ijkl}) = 0.5K[\phi_{ijkl} - \phi_0)]^2
+
+where the `K` is a constant, the angles should be provided in radians.
+The `N` is a multiplicity.
```
acfogarty: We can get rid of the reference to multiplicity here.

src/interaction/DihedralHarmonicNCos.hpp
```diff
@@ -202,7 +202,7 @@ namespace espressopp {
        * @return The value of the force
        */
       real _computeForceRaw(real phi) const {
-        return K * multiplicity * sin(phi0 - multiplicity*phi);
+        return -1.0*K * multiplicity * sin(phi0 - multiplicity*phi);
```
acfogarty: I agree about the minus sign, to make it -gradV, but like I said above I wonder if there's also a 1/sin(phi) term.

src/interaction/DihedralHarmonic.hpp
govarguz: Ok

src/interaction/DihedralHarmonic.py
govarguz: ok

src/interaction/DihedralHarmonic.hpp
govarguz: fully agree, got rid of it!

src/interaction/DihedralHarmonic.hpp
govarguz: Well, ForceRaw meant to me to treat the force as the potential is given, but now that you are commenting on it. And thanks for the detailed revision. I assume all the coordinates, also the raw ones must be handled in rectangular coordinates. Therefore 1/Sin(theta) should be applied. If there is any other feedback from @MrTheodor or @niktre, let me know.

MrTheodor: I've looked at TabulatedDihedral and I don't see any prefactor so we should include it in _computeForceRaw to be internally consistent. 
I am not sure about minus sign.. code is a bit obfuscated.

src/interaction/DihedralHarmonic.hpp
```diff
@@ -0,0 +1,201 @@
...
+          _phi = 1e-10; //not 0.0, because 1.0/sin(_phi) would cause a singularity
+        } else if (cos_phi < -1.0) {
+          cos_phi = -1.0;
+          _phi = M_PI-1e-10;
+        }
+
+        /// The part of the formula. 1/sin(phi) * d/dphi U(phi)
+        /// where the \f$U(\phi) = 0.5 K (\phi_{ijkn} - \phi_0)^2]\f$
+        ///
+        /// The derivative of \f$U(phi)\f$ is \f$K*sin(\phi_0 - \phi)\f$
```
MrTheodor: could you please fix the comment, there is no sin

src/interaction/DihedralHarmonic.hpp
```diff
@@ -0,0 +1,201 @@
...
+       * Compute raw force for the tabulated potential, depend only on the phi value
+       * \f$F(\phi) = -grad V = -grad [ 0.5 (\phi_{ijkn} - \phi_0)^2] ]  \f$
+       * where \f$phi_0\f$ is the reference angle
+       *
+       * @param[in] phi Degree in radians
+       *
+       * @return The value of the force
+       */
+      real _computeForceRaw(real phi) const {
+	real sin_phi = sin(phi);
```
MrTheodor: perhaps here we should also check if there will be 1/0 like in line 130

MrTheodor: Maybe it would be less error prone if `_computeForceRaw` would be directly used to compute coef1 in line 140.

src/tools/convert/gromacs.py
```diff
@@ -328,10 +327,10 @@ def read(gro_file, top_file="", doRegularExcl=True):
                 #is it really the dihedral (function type = 9) or is it actually the impropers (also labelled 'dihedraltypes' in gromacs but with function type = 4)
                 nextline = f.lines[lineindex + 2] #assumes one comment line between [ dihedraltypes ] and first parameters entry
                 nextline = nextline.split()
-                if (nextline[4]=='4'): 
+                if ((nextline[4]=='4') or (nextline[4]=='2')): 
                     readimptypes = True
                     readdhtypes = False
-                elif ((nextline[4]=='9') or (nextline[4]=='1') or (nextline[4]=='8')):
+                elif ((nextline[4]=='9') or (nextline[4]=='1') or (nextline[4]=='8') or (nextline[4]=='3')):
```
MrTheodor: Pythonic suggestion:
```python
elif int(nextline[4]) in [1, 3, 8, 9]:
```

src/interaction/DihedralHarmonic.hpp
govarguz: Sure!

src/tools/convert/gromacs.py
govarguz: good point!

src/interaction/DihedralHarmonic.hpp
acfogarty: Agreed about checking for 1/0, and in fact the checks should also use <= and not just <. I'll take care of it after merging Horacio's stuff.

junghans:
You shouldn't use the same commit message three times in row,  that is really not useful when digging through the history later.

govarguz:
Fully agree with you. However for this particular commit has being thoroughly reviewed and commented by several developers which simplifies going through this history and checking which corrections have been done. In particular with some signs and prefactors used in _ComputeForceRaw.

junghans:
Except that we only have the commit history and not the discussion from this issue 10 years down the road when espresso++ gets moved from github to the next hosting service.
image (Credits to [xkcd](https://xkcd.com/1296/))

PR #32, opened 2015-11-18
Fix pid output in XYZ C++ dump class

fedepad:
Very minimal changes (non disruptive) to fix bug in DumpXYZ.cpp .
Particle Id was not printed, just a ‘0’ for each particle was hard
coded as a pid.
We can think about making little changes in ConfigurationExt.cpp.
Now minimal changes to make it work.


vitst:
Well, this is a good question what is the standard .xyz format. In general it is
<number of atoms>
comment line
<element> <X> <Y> <Z>
where <element> is C, H, O, N etc. But usually people need more info.

A problem with PID as a first column appears when you try to visualize your .xyz using VMD, for example. You will get something like:
ERROR) BaseMolecule: Cannot add atom; namelist index value too large
This is for VMD for LINUXAMD64, version 1.9.1 (May be they changed it in 1.9.2).

Since we do not have elements but particles of different type, the best solution will be to replace "0" with type. I am not sure what was the reason to print just "0". May be ConfigurationExt did not have access to the type of the particle that time.

fedepad:
@govarguz just anticipated me. I think that as a simple hack we could just add a parameter in the DumpXYZ constructor, for example an int or a bool called viz. If viz True will dump a .xyz that makes VMD happy and used for visualization and if False will dump with PIDs so that eventually one could use for hypothetic restart and the user get actual useful information even to look for the given simulation as it appears we deal more with PIDs than types, otherwise type would not be always zero.
But for the type, as @vitst mentioned, instead of zero than the type should be made available through ConfigurationExt, if that info can be retrieved so that is not hard coded.

MrTheodor:
IMHO restarting simulation and writing a correct format of the trajectory are two different tasks and therefore should be splitted. I would rather add something in ``fastwriteXYZ`` if this is used only to restart like e.g. initialId = 0 and then use the index of loop as an atom id. 

btw, VMD will not read anything apart from type x y z:

**Unused Data**
``This plugin only reads the label and coordinates for each atom. Any additional data recorded for an atom is ignored.``
http://www.ks.uiuc.edu/Research/vmd/plugins/molfile/xyzplugin.html

govarguz:
@MrTheodor are you sure the PIDs are just the index of the loop while restarting a sim?
Concerning splitting them, sure they are to different tasks! Nonetheless, the difference will be exchanging a colum. Let's illustrate it, either XYZ format or Restarting format... But if the PIDs for restarting a sim are only the loop index, then we have a different story.

MrTheodor:
@govarguz of course depends how you label your particles, generally it is a wrong assumption so it is better to explicitly write id column but then it is no longer ```xyz format``` and should not be called in that way.
The problem with all of those flags is that they have some default values already set and you have to remember to switch it to true or false to get the right behaviour.  This switch has rather tremendous implication as it change the output format.

btw, are there two parallel discussions for this pull requests?

govarguz:
You have made a good point regarding splitting files. In particular if the ''DumpXYZflaggedVersion'' won't be well documented. Is there a writeRESTARTINGCONF somewhere in the package?

fedepad:
@MrTheodor  good point about two discussions, I think two things were mixed. Ok, since we do not have infinite time, let's keep the old DumpXYZ to conform to the "standard"; since from the source code of VMD and the page shown by @MrTheodor nothing is read after the coordinates, we can drop the velocities thus saving space and maybe time (roughly half space should be saved but probably less) as it makes no sense to keep them if it's just to visualize this file on VMD.
For what it concerns restart files, we need the pids, so I will write a separate class for that that includes the info needed for an eventual restart.
@govarguz As far as I see (but I might be wrong) there's no such thing in the package.

govarguz:
@junghans I would guess @fedepad has some updates on this. However I could not directly assign him could you do that?

junghans:
Done, I needed to add @fedepad as an external collaborator!

fedepad:
Ok, thanks @junghans .
@govarguz @MrTheodor @stuehn Last time here we agreed to separate the "strictly obey" to the XYZ format from eventual restart files (which probably will be in binary format). So I decided to fix the Dump XYZ format and print the type as first column which is not hard-coded anymore; now it queries the current pid and get the type for that pid.
I tried to do very minimal changes, of course more refined thinking can be thought to restructure some classes. This fix involves some changes in the Config* classes, as it was not exporting the type information before. 
So users of the DumpXYZ class should be able to print as usual (example polymer melt). Still comments in the code to keep the conversation. Please comment on the updated code. Notice type is size_t, while for example in the package mentioned by @pdebuyl , type might also be literal basically.

fedepad:
@junghans @MrTheodor I had to resolve merge conflicts otherwise Travis would not automatically pickup last commit from PR... https://github.com/travis-ci/travis-ci/issues/2099

src/io/DumpXYZ.hpp
```diff
@@ -49,11 +49,11 @@ namespace espressopp {
                         ParticleAccess(system), 
                         integrator(_integrator),
                         file_name( _file_name ),
-                        unfolded(_unfolded),
+                        unfolded(_unfolded),  // default should be unfolded right? where is this?
                         length_factor(_length_factor), 
                         append(_append){ 
         setLengthUnit(_length_unit);
-        if (system->comm->rank() == 0  && !append)
+        if (system->comm->rank() == 0  && !append) // why when is not append?
```
junghans: Are we clear here?

src/io/DumpXYZ.hpp
```diff
@@ -49,11 +49,11 @@ namespace espressopp {
                         ParticleAccess(system), 
                         integrator(_integrator),
                         file_name( _file_name ),
-                        unfolded(_unfolded),
+                        unfolded(_unfolded),  // default should be unfolded right? where is this?
```
junghans: Is it the default?

src/io/DumpXYZ.cpp
```diff
@@ -36,46 +37,82 @@ namespace espressopp {
...
             for(size_t i=0; i<num_of_particles; i++){
-            	//myfile << "bla" << endl;
-              myfile << "  0  " << length_factor * cei.nextProperties() << endl;
-              //myfile << "  0  " << length_factor * cei.nextCoordinates() << endl;
+              //myfile << "  0  " << length_factor * cei.nextProperties() << endl;
+              myfile << cei.Id() << " " << length_factor * cei.Properties() << endl;
+              cei.nextParticle();
             }
-          }
+          } */
```
junghans: Just remove the commented out code (see #39)!

src/analysis/ConfigurationExt.cpp
```diff
@@ -76,6 +86,13 @@ namespace espressopp {
     }
     */
 
+    // F.Padua: temp fix for better DumpXYZ support
```
junghans: Improve comment!

src/analysis/ConfigurationExt.cpp
```diff
@@ -55,6 +55,16 @@ namespace espressopp {
       //velocities[index] = Real3D(vx, vy, vz);
     }
 
+    void ConfigurationExt::set_id_vec_type(size_t index, RealND vec, size_t typ)
+        {
+          particleProperties[index].setDimension(vec.getDimension());
+          particleProperties[index] = vec;
+          particleType[index] = typ;
+          //coordinates[index] = Real3D(x, y, z);
+          //velocities[index] = Real3D(vx, vy, vz);
```
junghans: Remove commented out lines!

junghans:
@fedepad: Look good except minor pitpick, please also update the copyright year in the top of the touched files. Could you write a little test case and put it in `/testsuite`?

src/analysis/ConfigurationExt.hpp
```diff
@@ -1,5 +1,7 @@
 /*
-  Copyright (C) 2012-2015
+  Copyright (C) 2016
+      Max Planck Institute for Polymer Research & Johannes Gutenberg-Universität Mainz
+  Copyright (C) 2012,2013
```
junghans: What about 2014 and 2015?

testsuite/dumpXYZ/dumpXYZ.py
```diff
@@ -0,0 +1,42 @@
+#  Copyright (C) 2016
+#      Max Planck Institute for Polymer Research & Johannes
+#      Gutenberg-Universitaet Mainz
+#  Copyright (C) 2012,2013
+#      Max Planck Institute for Polymer Research
+#  Copyright (C) 2008,2009,2010,2011
+#      Max-Planck-Institute for Polymer Research & Fraunhofer SCAI
+#
```
junghans: This is a new file, so remove 2008-2015 copyright!

testsuite/dumpXYZ/run_test.sh
```diff
@@ -0,0 +1,22 @@
...
+#  REQUIREMENTS: ---
+#          BUGS: ---
+#         NOTES: ---
+#        AUTHOR: Federico G. Padua (), padua@uni-mainz.de
+#  ORGANIZATION: Johannes Gutenberg Universitat Mainz, ZDV
+#       CREATED: 02/11/2016 18:06
+#      REVISION:  ---
+#===============================================================================
+
+source ../../ESPRC
```
junghans: That won't work for out-of-source builds!

testsuite/dumpXYZ/dumpXYZ.py
```diff
@@ -0,0 +1,42 @@
...
+system.storage.addParticles([[5,23,espressopp.Real3D(4,1,2)]],'id','type','pos')
+system.storage.addParticles([[6,14,espressopp.Real3D(7,7,7)]],'id','type','pos')
+system.storage.addParticles([[7,6,espressopp.Real3D(9,9,9)]],'id','type','pos')
+system.storage.addParticles([[8,8,espressopp.Real3D(8,8,8)]],'id','type','pos')
+
+
+jack = espressopp.io.DumpXYZ(system, integrator, filename='test_dumpXYZ_type_not_hardcoded.xyz', unfolded = False, length_factor = 1.0, length_unit = 'LJ', append = False)
+jack.dump()
+
+print "Finished writing file, run the script to check for diff!"
```
junghans: Could you do the comparison in here? That would save us the shell script!

junghans:
The build will fail, I am working on that, see #56.

junghans:
Re-merge master and it should build fine.

src/analysis/ConfigurationExt.hpp
fedepad: I though about adding it, but @govarguz suggested me to put only the year of the commit I'm doing (2016 in this case). Thoughts on that? Let me know and I will do the changes if I need to add for example in this case: 2014,2015,2016.
I guess the same applies for example to src/analysis/ConfigurationsExt.cpp just to name one.

testsuite/dumpXYZ/dumpXYZ.py
fedepad: You're very right! Fixed!

testsuite/dumpXYZ/run_test.sh
fedepad: Removed bash script!

testsuite/dumpXYZ/dumpXYZ.py
fedepad: Done. Removed bash scripts.

src/analysis/ConfigurationExt.cpp
```diff
@@ -91,6 +109,58 @@ namespace espressopp {
...
+        boost::python::throw_error_already_set();
+      }
+
+      RealND props = (*it).second;
+      return props;
+    }
+
+
+    /*
+        Real3D ConfigurationExtIterator::nextCoordinates()
```
MrTheodor: why do you need this commented code?

src/analysis/ConfigurationExt.hpp
```diff
@@ -46,6 +49,11 @@ namespace espressopp {
 
       int nextId();
       const RealND nextProperties();
+      /* increment particle iterator, get pid, get properties (pos,v)   */
+      void nextParticle();
+      size_t Id();
+      const RealND Properties();
+
       //Real3D nextCoordinates();
       //Real3D nextVelocities();
```
MrTheodor: ```nextCoordinates``` is removed and I don't see any definition for ```nextVelocities();```

fedepad: Yes, it looks like that, again just kept as it was this part.
I guess previousely was started a nextCoordinates(), nextVelocities(), etc. and then opted for a single method Properties() and nextProperties().
Again, I can remove it and push again.

src/analysis/ConfigurationExt.cpp
fedepad: I guess you're referring to Real3D ConfigurationExtIterator::nextCoordinates(), but I see const RealND ConfigurationExtIterator::Properties() highlighted... 
Yes, Real3D ConfigurationExtIterator::nextCoordinates() is not used and probably the previous developers commented out thus just using const RealND ConfigurationExtIterator::Properties() for both coordinates and velocities.
Just kept as it was. We can think about removing this commented code...
Waiting feedback.

src/analysis/ConfigurationsExt.cpp
```diff
@@ -304,6 +306,209 @@ namespace espressopp {
...
+
+           // not master process, send data to master process
+
+           int tmp;
+
+           boost::mpi::status stat;
+
+           // wait for a signal (empty message) before sending
+
+           system.comm->irecv<int>(0, DEFAULT_TAG, tmp);
```
MrTheodor: you are using non-blocking receive, but then you don't care about the status, why?

src/analysis/ConfigurationsExt.cpp
```diff
@@ -304,6 +306,209 @@ namespace espressopp {
...
+          } else {
+
+           LOG4ESPP_INFO(logger, "proc " << system.comm->rank() << " sends data "
+                          << " of " << myN << " particles");
+
+           // not master process, send data to master process
+
+           int tmp;
+
+           boost::mpi::status stat;
```
MrTheodor: where do you use ```stat```?

src/analysis/ConfigurationsExt.cpp
```diff
@@ -304,6 +306,209 @@ namespace espressopp {
...
+                  req = system.comm->irecv<real>(iproc, DEFAULT_TAG, coordinates, 3*maxN);
+                  system.comm->send(iproc, DEFAULT_TAG, 0);
+                  stat = req.wait();
+                  nCoords = *stat.count<real>();
+
+                  // make sure to have 3 coordinate values for each id
+
+                  if (nCoords != 3 * nIds) {
+                    LOG4ESPP_ERROR(logger, "serious error collecting data, got " <<
+                                  nIds << " ids, but " << nCoords << " coordinates");
```
MrTheodor: shouldn't you stop the execution by throwing exception or MPI_Abort or exit(1), that serious error?

src/analysis/ConfigurationsExt.cpp
```diff
@@ -304,6 +306,209 @@ namespace espressopp {
...
+            }
+          }
+
+          if (i != myN) {
+            LOG4ESPP_ERROR(logger, "mismatch for number of local particles");
+          }
+
+          // each proc sends its data to master process
+
+          if (system.comm->rank() == 0) {
```
MrTheodor: I think that this part could be replaced by calling MPI_Gather method on all processes and collect on rank=0 results, then save it on rank=0 and at the end make synchronization by MPI_Barrier.

src/analysis/ConfigurationsExt.cpp
```diff
@@ -304,6 +306,209 @@ namespace espressopp {
       delete [] ids;
     }
 
+    void ConfigurationsExt::gather_modified() {
```
MrTheodor: why we need two methods, ```gather``` and ```gather_modified``` ?

src/io/DumpXYZ.py
```diff
@@ -104,12 +106,12 @@ class DumpXYZLocal(ParticleAccessLocal, io_DumpXYZ):
 
   def __init__(self, system, integrator, filename='out.xyz', unfolded=False, length_factor=1.0, length_unit='LJ', append=True):
     cxxinit(self, io_DumpXYZ, system, integrator, filename, unfolded, length_factor, length_unit, append)
-  
+
   def dump(self):
-    if not (pmi._PMIComm and pmi._PMIComm.isActive()) or pmi._MPIcomm.rank in pmi._PMIComm.getMPIcpugroup():
+    if not pmi._PMIComm or pmi._MPIcomm.rank in pmi._PMIComm.getMPIcpugroup():
```
MrTheodor:
```python 
if pmi.workerIsActive():
```

src/analysis/ConfigurationsExt.cpp
fedepad: Good point. Was done ok the first time is used, but not at the end. Can be modified along those lines.

src/analysis/ConfigurationsExt.cpp
fedepad: Currently not used. Can be added along the lines of similar code above in the same method.

src/analysis/ConfigurationsExt.cpp
fedepad: Can be done.

src/analysis/ConfigurationsExt.cpp
fedepad: That is the logic. And since I don't know if there are really performance improvement I wouldn't suggest to modify it. But we can think about that.

src/analysis/ConfigurationsExt.cpp
fedepad: We don't need two. I can remove one and rename it. I kept it when I wrote the modified version.

fedepad:
I'm closing this pull request as I'm in the process of modifying my Github. Will push from my branches the same code once I finish with this process.

junghans:
Where did these changes go?

fedepad:
@junghans My fault. I had to do major changes to my Github and repos and branches across repos. I have now a branch for this fix in my fork with the same changes Probably the less painful solution is to submit another PR and I finish to integrate the comments from here.
Otherwise if there is a sure/proved way to reopen this PR and make it pull the changes from a different branch of mine than the previous one that would be great! waiting for your suggestion/feedback!

junghans:
Ok, just refer to #32 somewhere in there.

PR #33, opened 2015-11-20
Fixed combination-rule 3 that was already in sigma epsilon units

govarguz:


govarguz:
yeah, PR is an entity and this one(in particular) has 2 different commits with different names on each commit. Whereas it is TRUE that 2 PRs could have worked better.

junghans:
Add 2015 in the header, otherwise looks fine!

src/tools/convert/gromacs.py
```diff
@@ -1,4 +1,4 @@
-#  Copyright (C) 2012,2013
+#  Copyright (C) 2012,2013,2015(H)
```
junghans: What is 2015(H)?

Issue #37, opened 2015-12-09
running ESPP in Garching


niktre:
Hi, 
I've managed to compile ESPP in Garching on their HPC. To do so, I did:
```
module load git/2.1
module load cmake/3.2
module load fftw/gcc/3.3.4   
module load gcc/4.9

cmake -DEXTERNAL_BOOST=OFF -DEXTERNAL_MPI4PY=OFF -DFFTW3_INCLUDES=$FFTW_HOME/include -DFFTW3_LIBRARIES=$FFTW_HOME/lib/libfftw3.a .
```

The problem is, that when I'm starting the script, upon importing espressopp, I get:
```
SyntaxError: invalid syntax
Traceback (most recent call last):
  File "start_lb.py", line 3, in <module>
    import espressopp
  File "/ptmp/nikt/parall_espressopp/espressopp/__init__.py", line 65, in <module>
    from espressopp import esutil, bc, storage, integrator, interaction, analysis, tools, standard_system, external, check, io
  File "/ptmp/nikt/parall_espressopp/espressopp/storage/__init__.py", line 26, in <module>
    from espressopp.storage.DomainDecomposition import *
  File "/ptmp/nikt/parall_espressopp/espressopp/storage/DomainDecomposition.py", line 49, in <module>
    from espressopp.tools import decomp
  File "/ptmp/nikt/parall_espressopp/espressopp/tools/__init__.py", line 31, in <module>
    from espressopp.tools.convert import *
  File "/ptmp/nikt/parall_espressopp/espressopp/tools/convert/__init__.py", line 23, in <module>
    import gromacs
  File "/ptmp/nikt/parall_espressopp/espressopp/tools/convert/gromacs.py", line 25, in <module>
    from topology_helper import *
  File "/ptmp/nikt/parall_espressopp/espressopp/tools/convert/topology_helper.py", line 177
    pot = espressopp.interaction.DihedralRB(**{k: v for k, v in self.parameters.iteritems()})
```

Does anyone have an idea if this syntax in ```topology_helper.py``` is fine?

niktre:
Sorry for several (un-)assignments

MrTheodor:
Hi,

Which version of Python do you use? Is it 2.6.x ? That could be the reason
besides that I don't remember why this dictionary comprehension is there.
I will check if it is required.

Best,
Jakub

niktre:
[...]

PR #38, opened 2015-12-09
removed dict comprehension

MrTheodor:
It should fix issue #37 
cc: @niktre


junghans:
Looks great, please update copyright, too!

PR #40, opened 2015-12-23
Fixed SystemMonitor in case of parallel run.

MrTheodor:
Call to gather-all was not done on all cpus. Still it seems to be suboptimal as there is no point to have available the total calculations of observables on all cpus.

*minor*: removed calculation of a temperature and a kinetic energy in the SystemMonitor, it should be done
as other observables.

### DIFF
--- a/src/analysis/SystemMonitor.cpp
+++ b/src/analysis/SystemMonitor.cpp
@@ -29,36 +29,20 @@ namespace espressopp {
 namespace analysis {
 
 void SystemMonitor::perform_action() {
+  current_step_ = integrator_->getStep();
+  values_->clear();
+  values_->push_back(current_step_);
+  values_->push_back(current_step_ * integrator_->getTimeStep());
+
+  computeObservables();
   if (system_->comm->rank() == 0) {
-    current_step_ = integrator_->getStep();
-    values_->clear();
-    values_->push_back(current_step_);
-    values_->push_back(current_step_ * integrator_->getTimeStep());
-
-    computeKineticEnergy();
-    computeObservables();
-
-    write();
+    output_->write();
   }
 }
 
 void SystemMonitor::computeObservables() {
   for (ObservableList::iterator it = observables_.begin(); it != observables_.end(); ++it) {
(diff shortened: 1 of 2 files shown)


MrTheodor:
any progress on that?

govarguz:
I am not sure if I follow the reason of SystemMonitor now, besides of being at c++ level. So far, this patch enables each MPI task for keeping the observable(s) of interest, on a each local domain (core) basis. 
Could you explain more in detail on the example you provided, by highlighting the difference with the existing (standard) analysis tools?
BTW, I mixed this bug with an issue. The labels were confusing since this is the first PR I see with a BUG label on it...

MrTheodor:
@govarguz That's nothing to do with any analysis tools. It's rather a manager that one can use to add observables with some names, then push whole ```SystemMonitor``` object to ```ExtAnalysis``` and forget about it. It will compute observables and stores values to the csv file. The example is in the documentation.

And this patch, well silly mistake, on each of CPU ```compute``` has to be called because of ```all_reduce``` methods in the e.g. ```Temperature.hpp```.

govarguz:
@MrTheodor as you pointed out...the difference between using ```ExtAnalysis``` alone and pushing the ```SystemMonitor``` object to ```ExtAnalysis```. I meant highlighting this kind of versatility/differences in the example you gave. Extending  the example will be worthy. 
Another suggestion, nowadays it will be better using the word CORE than CPU.

MrTheodor:
@govarguz sure nothing special is happening here, except from delegating work to one routing and not bother of any additional code to store values. Perhaps it would be nice to define different frequencies of computing values for added observables, but I didn't need that. 
IMHO example in the documentation shows a simple case how to use it.

I would rather say process ;-) anyway..

govarguz:
@MrTheodor, would adding different frequencies increase considerably performance? 
I'd suggest to add the frequencies would improve this PR and SysMonitor.

MrTheodor:
@govarguz good idea, but that has to be integrated with other file format, perhaps HDF5; 
I will create an issue as this one is only to fix the bug

PR #41, opened 2015-12-24
Added support for naming interaction when defined by System.addInteraction

MrTheodor:
Additional name for the interactions that are defined in System. By this it is possible to remove
interactions with the cleaner way by the name:
``` python
system.removeInteractionByName('lj')
```

The API does not change, it is still possible to define interaction without the name.


govarguz:
Looks good to me. 
cc. @stuehn

Issue #42, opened 2016-01-02
http://www.espresso-pp.de/ does not work


MrTheodor:
Web page does not work (404). Also ping on espresso-pp.de return ```ping: unknown host espresso-pp.de```

Perhaps the manual could be hosted on ```https://readthedocs.org/``` and the home page on github.io ?

junghans:
see #43

govarguz:
web page is up

PR #43, opened 2016-01-20
Auto-deploy sphinx

junghans:
Use new rule `make ug` to build user guide using sphinx and upload it to http://espressopp.github.io/ automatically for every push to master (not on pull requests though).


junghans:
@MrTheodor : any comments?

MrTheodor:
Great, working like a hell! 

With travis, does every push will also update github.io ?

On http://espressopp.github.io/ the left menu does not fit very well perhaps because of some missing css files; anyway if the line conf.py:52 ```templates_path = ['_templates'] is commented out then everything looks better.

junghans:
Yes, every push to `master` will create an new commit to github.io. Other branches and pull requests are excluded.

@MrTheodor : Can you do that css fix in another pull request!

MrTheodor:
Now I see the problem. The documentation uses ```autodoc``` extension; so the documentation will not be properly generated without sourcing ESPRC.

CMakeLists.txt
```diff
@@ -197,3 +197,27 @@ else ( DOXYGEN_FOUND )
...
+  NAMES sphinx-build
+  DOC "Sphinx documentation generation tool (http://www.sphinx-doc.org/)"
+)
+
+if ( SPHINX_EXECUTABLE )
+
+   set ( SPHINX_BUILD_ROOT "${CMAKE_SOURCE_DIR}/doc/ug" )
+
+   add_custom_target(ug
+      COMMAND ${SPHINX_EXECUTABLE} -b html . _build/html
```
MrTheodor: This should help to handle the ```autodoc``` sphinx extension out-of-box:
```CMake
COMMAND export PYTHONPATH="\$PYTHONPATH:${CMAKE_BINARY_DIR}" && ${SPHINX_EXECUTABLE} -b html . _build/html
```

junghans: Also see: https://cmake.org/Bug/view.php?id=14306

MrTheodor: Nice solution!

junghans: >cmake-3.1 comes with the `env` command using that now.

junghans:
I included #44 in this pull request, looks much better now.

@stuehn : any comments?

acfogarty:
There are also a few other problems with http://espressopp.github.io/ : missing most of the html pages linked to from the left menu bar (all carried over from www.espresso-pp.de). I'm looking into it now...

MrTheodor:
@acfogarty that is because ```_templates/layout.html``` contains those links
perhaps content of those static pages could be put in .rst files (that will be a mix of homepage and user documentation)

junghans:
@acfogarty http://espressopp.github.io/ is out-of-date this pull request doesn't update it, I just did that manually, so that we can see how it currently looks. Running `git grep espresso-pp.de` on espressopp/espressopp.github.io@7eb8db1d7582ed204b0a45b1dbbd1fe8e72650e9 returned nothing, so all these broken links should be gone.

On a related note to what @MrTheodor said, we could generate build the whole homepage using travis, which would fix #42 once and for all.

MrTheodor:
@junghans @acfogarty 

I've update #44 and included data from the homepage.

acfogarty:
@junghans @MrTheodor Okay thanks, now I understand.

We need to keep _templates/layout.html (re:  #44 ) because @stuehn is using it when generating documentation to insert into www.espresso-pp.de.

As well as the espressopp.github.io documentation for users using the developer version, we need to keep a separate, not-automatically generated documentation page at www.espresso-pp.de (notwithstanding #42 ) because this is the documentation corresponding to the stable release version of espressopp (i.e. the tarball at http://www.espresso-pp.de/Download.html)

So we need to be able to generate both documentations from the same doc/ug source. Can we keep _templates/layout.html and have a flag not using that line in conf.py when building from travis? (I tried to figure out how to do that myself but can't.)

As for the other static pages of www.espresso-pp.de (Downloads, Impressum, Developer Team etc.) it looks to me like @MrTheodor has already included the only necessary one as getting_help.rst (#44) and everything else is not relevant for the espressopp.github.io version of the documentation.

MrTheodor:
@acfogarty check now #44 , I've forgot to commit other files..

@stuehn for the releases, why not to use https://github.com/espressopp/espressopp/releases ?

IMHO it should be possible to have two conf.py or to pass some environmental variable, it's nothing more than Python file.

junghans:
@acfogarty : One could make a second `conf.py` to build the homepage or check for an environment variable (e.g. `BUILD_HOMEPAGE`) inside `conf.py` to select the build type.

.travis.yml
```diff
@@ -6,18 +6,35 @@ before_install:
...
+        git config --global user.email "espp-devel@listserv.mpip-mainz.mpg.de";
+        git add --all && git commit -m "Documentation Update";
+        openssl aes-256-cbc -K $encrypted_194b3d1e9306_key -iv $encrypted_194b3d1e9306_iv -in ../../deploy.enc -out ~/.ssh/id_rsa -d;
+        chmod 600 ~/.ssh/id_rsa;
+        git push git@github.com:espressopp/espressopp.github.io.git master;
+      else
+        git diff --no-color;
+      fi;
+    fi
+
```
junghans: @certik: FYI

certik: @junghans thanks! Very useful.

junghans:
#44 is not included anymore, so we should merge this and then work on the layout.

acfogarty:
@junghans Okay. I wanted to fix the layout first (it seems you can't pass your own variables/flags to sphinx-build from the command line but you can specify different conf.py files) but I should do that in a separate pull request.

junghans:
Ok, auto-deployment worked and http://espressopp.github.io/ doesn't look bad.

PR #44, opened 2016-01-22
Updated conf.py for sphinx and removed layout.html

MrTheodor:
Removed layout.html, it looks like as a template for home page but there are no other files.

updated conf.py to fetch automaticly version of ESPP from Version() module

copied content of http://www.espresso-pp.de/

cc: @stuehn @junghans


doc/ug/conf.py
```diff
@@ -12,6 +12,16 @@
 # serve to show the default.
 
 import sys, os
+import datetime
+
+# Gets version directly from the code
+try:
+    import espressopp
```
junghans: Not sure if that will even use the current version or the one which is installed.
I might be better to turn `conf.py` into a cmake configure_file and replace the version at cmake time.

MrTheodor: Would it be possible? The version is hardcoded in Version.hpp so some sort of parsing would be required. Other thing is that, during the building of documentation, ```espressopp``` module has to be available as ```autodoc``` depends on it.

junghans: I see, in #43 we could make the `ug` target depend on the espresso++ model and set the environment correctly like you suggested.

MrTheodor:
Included in #43

junghans:
Leave it open until #43 is merged.

junghans:
I think we should leave the layout like it is and make http://espressopp.github.io/ just a copy of http://www.espresso-pp.de/ with more recent content on it.

It shouldn't be hard to generate and auto-deploy the [pdf manual](https://github.com/espressopp/espressopp/blob/master/espp_manual.pdf) as well.

MrTheodor:
Working on that.

cc: @acfogarty

PR #52, opened 2016-02-08
DumpH5MD and DumpTopology

MrTheodor:
Two extensions:
- `DumpH5MD` that stores information about particles in [HDF5](https://www.hdfgroup.org/HDF5/) file using [H5MD](http://nongnu.org/h5md/) format. This is mostly based on the work done by @pdebuyl here: https://github.com/pdebuyl/espressopp/tree/dev_PyStore
- `DumpTopology` that stores pairs of bonds from `FixedPairList` in HDF5 file.
- `Particle` - added new field `res_id` that is an id of whole molecule (like single polymer chain)

minor things in `System` and `FixedPairList`.

cc: @govarguz


junghans:
Should we install hdf5 in travis as well?

govarguz:
@MrTheodor the extensions are valid, however there are current not committed efforts for dumping files (also hdf5 and perhaps h5md) in single and parallel I/O modes with their corresponding dumping files timing (and some benchmarks) cc: @fedepad . Therefore @fedepad feedback on this PR can be also a plus.

MrTheodor:
@govarguz as soon as the underlying gpfs, libhdf5 and h5py would correctly work with parallel writing, this dump will do the same. It's entirely dependent on h5py library. The same is with DumpTopology. The only problem with ```DumpH5MD``` is that it's not ready to work with ```ExtAnalyze```.

fedepad:
@junghans I think it's a good idea as I see @MrTheodor needs it for this pull request and if someone else is developing something in that direction would be useful to have it installed. If doesn't take much effort on your side would be nice to have it there.
@MrTheodor when you say it's entirely dependent on h5py what do you really mean? Do you mean that without installing h5py is not going to work? For anything, I would only rely on h5py just for some preliminary tests at the python level to see timings and understand the how to write the info (that's what I tried I while ago for example), but it's still behind the C interface and the parallel version of h5py has quirks. Plus is slower than the C interface although might be easier to code only at the python level.
Another part I do not understand is when you say "as soon as the underlying gpfs, libhdf5 and h5py would correctly work with parallel writing"; the way it's written looks like at the moment GPFS and libhdf5 do not work correctly in this case, so as soon as they are fixed everything it's fine. I'm pretty sure parallel writing has been used in the combination GPFS and HDF5, which in its truly parallel part relies on MPI-IO but I might be wrong. That this part is rarely used or there are performance concerns that's another point. Maybe you meant something else...let me know!

MrTheodor:
@fedepad: Yes, it's required h5py and ```DumpTopology``` required also pyh5md library. 
What I had in mind is that hdf5 and h5py have to be compiled with support for MPI and also on your cluster GPFS has to be installed and configured properly.  @pdebuyl did nice work with those py_buffers which do not required to copy data from C++ to Python code, so at least at this point it's not so bad.
@junghans currently not, I haven't put any testsuite.

junghans:
@MrTheodor , write a test case and I will add the missing part to `.travis.yml`.

MrTheodor:
I did some benchmarking on writing with this DumpH5MD compared to DumpGRO and to case where no trajectory is stored.

abct

description:
https://www.dropbox.com/s/1wj2k3h4dzwdy6i/DumpH5MD%20benchmark.pdf?dl=0

scripts:
https://dl.dropboxusercontent.com/u/525038/h5md-benchmark-04-05-2016.tar.gz

cc: @fedepad @govarguz

pdebuyl:
Hi Jakub,

On both graphs, higher values indicate faster execution. So, DumpH5MD is actually faster for single-node execution?

MrTheodor:
@pdebuyl From the raw data (included in tar.gz), for single node, the execution time are 4.63e+03 (H5MD) and 4.59e+03 (pure simulation without traj.) so H5MD is slower a bit (for two simulation, the difference is at third digit) comparing to a simulation without storing trajectory.

fedepad:
@MrTheodor Thanks for this benchmarks. I have done "similar" benchmarks for parallel writing shared HDF5 file with all the particle info for about 2.5M particles of a polymer melt simulation. Since it's not that clear to me from the numbers, I would like to know few things:
1) how much was the total time spent in writing (I/O, sec) and what is the total time of the simulation; and how much time to write all the particle info (in sec) in a single time step?
2) It seems you show the average of three runs per simulation (looking at the run scripts...). Is that correct?
3) Where are you dumping these files? It's a node-local storage, or is it a networked file system like NFS or is it a parallel FS like Lustre or GPFS for example? This is important.
 
And I think @pdebuyl was probably referring in comparison to DumpGRO, not to the pure LJ simulation. There's no way a simulation with file writing (DumpH5MD or DumpGRO) can be faster (same parameters, same conditions, same load) than one where no file writing  is happening, otherwise measurements should be reviewed...

pdebuyl:
@fedepad This was the idea, yes.

MrTheodor:
> 1) how much was the total time spent in writing (I/O, sec) and what is the total time of the simulation; and how much time to write all the particle info (in sec) in a single time step?

I haven't measure this yet, however the chunk size matter because the resize of the data is quite expensive.
I have added some new timers to measure more precisely which part is the slowest. 

> 2) It seems you show the average of three runs per simulation (looking at the run scripts...). Is that correct?

Yes, I run three times. 

> 3) Where are you dumping these files? It's a node-local storage, or is it a networked file system like NFS or is it a parallel FS like Lustre or GPFS for example? This is important.

This is a GPFS file system that is shared with computing nodes (ThinKing and storage on DDN1). Here is the hardware specification https://www.vscentrum.be/infrastructure/hardware/hardware-kul

govarguz:
ping @fedepad

fedepad:
@MrTheodor I'm back, I will checkout your branch and ran at different scale on our cluster for some measurements. Can you leave me your email or some chat handle like Jabber, irc, Skype or other for faster communication? Or if you have slack (which we could setup for E++) or Gitter which one can use logging in with the Github account?

src/io/DumpH5MD.cpp
```diff
@@ -0,0 +1,332 @@
...
+      i++;
+    }
+  }
+}
+
+PyObject* DumpH5MD::getPosition() {
+  if (store_position && position.len) return PyMemoryView_FromBuffer(&position);
+  Py_INCREF(Py_None);
+  return Py_None;
+}
```
fedepad: Add blank line between methods as the other methods below.

MrTheodor: Done

MrTheodor:
@fedepad sure: jkrajniak@gmail.com -> hangout

fedepad:
@MrTheodor FYI: I finished the setup, which wasn't comfortable to put it mildly since there was no python shared lib, just a static one, so required some hacking on the produced CMake files, plus other stuff regarding mpi,etc. But I think it's ok now and some very small checks with the polymer melt seems to run, so probably tomorrow I will try to setup some scripts for the measurements with bigger simul (~2.5M particles), and some other sanity checks before I start some runs in the next days after I have the ok of the sys admins that the cluster is fine. I will get back to you privately.

fedepad:
It's about the do_sort flag.
I know it's written in the DumpH5MD.py, but it feels somehow hidden for a user a complete working example in that case.
It works if close() is explicitly called otherwise, even activating the flag, doesn't work (I know it's in the documentation, but a small example I think would be better, just adding the close call...).

Because when used in this way (which going fast through the doc might happen):

dump_h5md = espressopp.io.DumpH5MD(system, filename="test_small.h5", do_sort=True, integrator=integrator)
dump_h5md.dump()

which is what is usually done with the other IO classes but doesn't work for sorting even if the flag is on; so the small working example, if one would like the file sorted would be in this case:

dump_h5md = espressopp.io.DumpH5MD(system, filename="test_small.h5", do_sort=True, integrator=integrator)
dump_h5md.dump()
dump_h5md.close()

Maybe you can add a working example like this when mentioning the do_sort flag in DumpH5MD.py, just it's clear 100% its use.

BTW I understand you are separating "opening", from dumping and closing, since I bet I'm going through similar troubles...

fedepad:
@MrTheodor Update: Currently running on GPFS. I will try to run on Lustre too if setup for E++ build will be fast on the other cluster.

pdebuyl:
Hi @fedepad @MrTheodor 
There has been a lot of activity here. I'd be glad to test the dump system for H5MD. What branch should I clone for testing? MrTheodor/espressopp/DumpH5MD or one from @fedepad's account?
I have only access to the same system as @MrTheodor but I'd be glad to have a look at the resulting files.

fedepad:
Hi @pdebuyl thanks for your help and for you availability for checking those!
So my idea was just to test both for performance at bigger scale and then pick one to merge or something or having it in a separate branch so that we have faster I/O in the tree (see #77 ). 
Basically I run the Polymer melt example with at least 2.56M or 5M particles and see the results for what it concern raw performance and in my case if the MPI hints have a visible effect.
I can push my "version" in the late afternoon to my fork and notify you and give the link, so that you can have a look at the files produced if they conform to the standard, etc...

pdebuyl:
Hi @fedepad, I'll have a look either tomorrow or next week then. What is the typical core count / CPU time for the benchmarks?

fedepad:
@pdebuyl You should be able to have a working version from here:
https://github.com/fedepad/espressopp/tree/io-h5md
Since I've done some cleaning and removed files maybe I've missed something but you should be able to build it ok, given a parallel version of hdf5. Mind doc isn't updated.
For the cores count, etc. I'm running with 64 up to 512 cores at least.
I've done few tests now, but @MrTheodor version was fast. I think writing pids and positions with 64 cores of ~2.56M particles time.clock() was reporting to take around 1 sec a couple of days ago.
For example yesterday when I was trying with 256 cores ~5M particles there were some problems, which might due to the mpi implementation. For this reason I'll try also another mpi package for both versions.
I will run more times, I will produce a graph for our reference.
My version I'm trying to get it to build there, since I need to trick the Cmake generated files (same I had to do with Jakub version when I setup his stuff).

fedepad:
@MrTheodor Grabbing your last push!

src/io/DumpH5MD.py
```diff
@@ -0,0 +1,483 @@
...
+        else:
+            self.integrator = system.integrator
+
+        if self.integrator is None:
+            raise RuntimeError('Integrator is not set properly.')
+
+        self.file = pyh5md.H5MD_File(filename, 'w', driver='mpio', comm=MPI.COMM_WORLD,
+                                     creator='espressopp',
+                                     creator_version=espressopp.VersionLocal().info(),
+                                     author=author, email=email
```
fedepad: there's no "email" field in pyh5md.H5MD_File and for this reason no email is printed in the resulting H5MD file. Please change email to author_email, otherwise doesn't get populated.

fedepad: To be precise it tries to populate the email attribute but it will not because it's looking for an argument named author_email to populate the attribute email.

fedepad:
@pdebuyl please grap the updated version with respect to the last one, still hosted in the same place:
https://github.com/fedepad/espressopp/tree/io-h5md

fedepad:
Hi guys, finally a graph, sorry I had to take care of other stuff too.

[fede_jakub_H5MD_runs_HPC_cluster_Mainz.pdf](https://github.com/espressopp/espressopp/files/614917/fede_jakub_H5MD_runs_HPC_cluster_Mainz.pdf)


The graph show the time as experienced by the user (with the python time.clock() between the points of interest, i.e. the constructor of the class, the dump routine and the close routine), i.e:
start = time.clock()
Constructor()
end = time.clock

and so on, you got the point.
Bar on the left is mine (if that's not super clear from the colors...sorry!), bar on the left Jakub's.

The simulation that was run was the polymer melt example with 5M particles and only one timestep, just to register the time taken with both approaches. Every configuration (given number of MPI ranks and given implementation (me, Jakub)) was run 10 times and average and std are shown on the graph.

Between 64 --> 128 you see a bit of a bump in time and that's because every compute node in this cluster has 64 cores and so as soon as you go over 64 MPI ranks, inter-node communication and coordination (with related bigger latencies) start to kick in.

Notice that the same chunk size was used, i.e. 256, as the default 128 in Jakub's code was giving worst results. The particle data recorded is: pid, position, mass, image.

I will put at some point (......) the data in archives and share it...

MrTheodor:
@fedepad interesting! which method of writing in your code did you test, was it N-to-1 ?

fedepad:
@MrTheodor Yes, it's N-to-1. Mind it's H5MD as yours, it's the one I mentioned just before my last post in this thread, not the one for which I opened the pull request a while ago, just to keep them separate at the moment...
It can also do N-to-N just switching a parameter in the python constructor, but that one is way faster than this as you basically avoid all the extra communication, but here we mentioned to test N-to-1 for the moment.

MrTheodor:
@fedepad what is the difference for single node between pure C++ and h5py code? I would expect that C++ should be faster as there is no overhead of Python layer. Interesting is the multi-node case. I suppose that this is because Python version does not use collective write. 
Recently Pierre update his pyh5md library to enable this, I will update the branch with those changes.
Could you share the scripts that you have used for benchmark?, I would like to run it on our cluster.

fedepad:
Attached the scripts used (you mean the polymer melt I guess, not the job scripts...)...sorry I kept it open here but didn't click "comment" ;) ...
Yes, probably the new use of collectives in pyh5md will bring things closer for multi-node runs to my version, usually is the stuff that makes a big difference.
For the single node, I honestly don't understand why the h5py version looks faster and what might be the causes, as my code already uses bitflags for selecting the data to store and avoids extra comparisons in the main particle loop. 
Many (many) months ago I was trying out writing simple stuff with h5py and it appeared slower but probably they have improved stuff a lot.
I use ch5md with some modifications to allow parallel stuff, but maybe can be improved a bit by changing how this stuff is done within ch5md.
Also, with the new HDF5 and probably the addition of routines that allow to write multiple datasets in one call, there might be chances that changing that part in ch5md to accommodate these calls might bring even further improvements as mentioned by HDF group in one post I don't remember...

[scripts_used.tar.gz](https://github.com/espressopp/espressopp/files/619279/scripts_used.tar.gz)

govarguz:
@junghans @niktre @stuehn LGTM, with the following BUTs

1. I am in favor of using either h5py or h5xx for this task (any low level calls of HDF5 calls from @fedepad can be used for specific cases, or Computer Science benchmark development, however not there for the stable E++ version.
2. In the current PR h5py cannot be used it with ExtAnalyze extension. Nonetheless it can be called explicitly from the Python script call .dump() method. This I would consider as a task to do in the future. It should be possible as it is possible to call python methods from the c++ code via boost. @MrTheodor This should be added in the documentation and also as a milestone.

@MrTheodor Finally I also noticed in several files that the copyright header of the E++ project is not included, can you add it? and also year 2018

MrTheodor:
@govarguz updated doc and files. Regarding this point:

> @MrTheodor Finally I also noticed in several files that the copyright header of the E++ project is not included, can you add it? and also year 2018

Do you referee to the files in `src/io/pyh5md` ? Those are actually from the @pdebuyl repository so I did not want (not sure even if I am allowed to do this) to put E++ header.

govarguz:
@MrTheodor @pdebuyl  no need to use the header for such files. and the rest LGTM, the issues have been also created

PR #53, opened 2016-02-09
Add support for code coverage

junghans:
See:
<https://codecov.io/github/junghans/espressopp?branch=coverage>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,19 +23,26 @@ before_install:
  - ln -s /usr/bin/ccache "$HOME/bin/clang++"
  - ln -s /usr/bin/ccache "$HOME/bin/clang"
  - if [[ ${EXTERNAL} = ON ]]; then pip install --user mpi4py==1.3.1; fi 
+ - if [[ ${COVERAGE}  ]]; then pip install --user codecov; fi 
 
 env: #maybe add mpich later
   global:
     - CCACHE_CPP2=yes
   matrix:
     - EXTERNAL=ON BUILD_UG=ON CMAKE_VERSION=3.1.3-Linux-x86_64
+    - EXTERNAL=ON COVERAGE=ON
     - EXTERNAL=OFF
+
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -23,19 +23,26 @@ before_install:
...
+    - compiler: clang
+      env: EXTERNAL=ON COVERAGE=ON #clang has no coverage support 
+
 script:
   - mkdir build && cd build && 
-    PATH="$HOME/bin:/usr/lib/ccache:$PATH" cmake -DEXTERNAL_BOOST=$EXTERNAL -DEXTERNAL_MPI4PY=$EXTERNAL .. && 
+    PATH="$HOME/bin:/usr/lib/ccache:$PATH" cmake -DEXTERNAL_BOOST=$EXTERNAL -DEXTERNAL_MPI4PY=$EXTERNAL ${COVERAGE:+-DUSE_GCOV=ON} .. && 
     make -j4 all ${BUILD_UG:+ug} &&
-    make test ARGS="-V" &&
+    make test ARGS="-V ${COVERAGE:+ -E pi_water}" &&
```
MrTheodor: why?

MrTheodor:
I was thinking about this tool, Espresso has it for a while. But is it really deal correctly with the templates, which are the core of ESPP ?

.travis.yml
junghans: The `pi_water` test runs too long with `-O0`, I will fix that in another pull request.

junghans:
Sure codecov isn't perfect, but better than nothing and in addition we can upload the same data to other services like coverall.io as well, all of them use `gcov` as a starting point.

.travis.yml
junghans: See: https://travis-ci.org/junghans/espressopp/jobs/107892462

junghans: See #54

PR #55, opened 2016-02-11
Documentation github.io page

MrTheodor:
Updated documentation layout
moved most parts of espresso-pp.de to .rst files; only missing things is
a list of publications, instead there is a link to Google Scholar.

fixed some errors that sphinx pointed out in .py and .rst files
includes MrTheodor/espressopp/pull/1

cc: @stuehn @acfogarty


src/analysis/OrderParameter.py
```diff
@@ -24,32 +24,6 @@
...
-if pmi.isController :
-    class OrderParameter(AnalysisBase):
-        __metaclass__ = pmi.Proxy
-        pmiproxydefs = dict(
-          cls =  'espressopp.analysis.OrderParameterLocal',
-          pmiproperty = [ 'cutoff', 'l' ]
-        )
-
-
-
```
junghans: Why are you removing these lines?

junghans:
@MrTheodor : looks great, just one question!

src/analysis/OrderParameter.py
MrTheodor: Is it a mistake of some auto generator of documentation? The code bellow is very similar to the one that is in documentation section.

PR #60, opened 2016-02-17
Hdf5 file writer, with test scripts for timing and cluster, to review.

fedepad:
Here for review small class that I created while ago, writer HDF5 file, no fancy; N-to-1, N-to-N I/O mode and MPI hint specification/optimization for GPFS blocksize on cluster.
Tests have been running for two days. The "default" polymer melt tested with both writers up to 512/1024 cores on GPFS.
Main CmakeList probably will not work as parallel HDF5 was installed somehow so the else branch in HDF5 hasn't been tested (if not hdf5 fetch it and build it);
On multiple clusters at different HPC centers experienced problems with Cmake recognizing the loaded boost in the environment. Had to use the one in contrib/ to be able to run.


CMakeLists.txt
```diff
@@ -160,6 +161,76 @@ else()
...
+        set(ExternalProjectCMakeArgs        
+            -DBUILD_SHARED_LIBS=ON
+            -DHDF5_BUILD_HL_LIB=ON
+            -DHDF5_BUILD_TOOLS=ON
+            -DHDF5_ENABLE_PARALLEL=ON
+            )
+        set(_hdf5_version "1.8.15-patch1")
+        file(MAKE_DIRECTORY ${CMAKE_SOURCE_DIR}/contrib/hdf5_local/src)
+        ExternalProject_Add(hdf5_local
+            DOWNLOAD_COMMAND curl http://www.hdfgroup.org/ftp/HDF5/current/src/hdf5-${_hdf5_version}.tar.gz | tar xvz -C ${CMAKE_SOURCE_DIR}/contrib/hdf5_local/src
```
junghans: Is there a reason why you customize that? cmake has a build-in downloader!

CMakeLists.txt
```diff
@@ -160,6 +161,76 @@ else()
...
+            CMAKE_ARGS ${ExternalProjectCMakeArgs}
+            SOURCE_DIR ${CMAKE_SOURCE_DIR}/contrib/hdf5_local/src/hdf5-${_hdf5_version}
+            BINARY_DIR ${CMAKE_SOURCE_DIR}/contrib/hdf5_local/build
+            STAMP_DIR  ${CMAKE_SOURCE_DIR}/contrib/hdf5_local/stamp
+            TMP_DIR    ${CMAKE_SOURCE_DIR}/contrib/hdf5_local/tmp
```
junghans: Why do you need to customize these?

CMakeLists.txt
```diff
@@ -160,6 +161,76 @@ else()
...
+            #${CMAKE_SOURCE_DIR}/contrib/hdf5_local/src/hdf5-${_hdf5_version}/c++/src
+            ${CMAKE_SOURCE_DIR}/contrib/hdf5_local/build/include
+            )
+        set(_hdf5_libs
+            ${CMAKE_SOURCE_DIR}/contrib/hdf5_local/build/lib/libhdf5_hl.so
+            ${CMAKE_SOURCE_DIR}/contrib/hdf5_local/build/lib/libhdf5.so
+            -ldl
+            )
+        add_definitions( -DHDF5_LAYER )
+    endif()
```
junghans: just use empty `HDF5_LIBRARIES`.
```
else()
set(HDF5_LIBRARIES)
```

src/CMakeLists.txt
```diff
@@ -42,7 +50,13 @@ endif (WITH_RC_FILES)
...
 
 add_library(_espressopp ${ESPRESSO_SOURCES})
-target_link_libraries(_espressopp ${Boost_LIBRARIES} ${PYTHON_LIBRARIES} ${MPI_LIBRARIES} ${FFTW3_LIBRARIES} ${VAMPIRTRACE_LIBRARIES})
+
+if(NOT ENABLE_HDF5)
+    target_link_libraries(_espressopp ${Boost_LIBRARIES} ${PYTHON_LIBRARIES} ${MPI_LIBRARIES} ${FFTW3_LIBRARIES} ${VAMPIRTRACE_LIBRARIES})
+else()
+    target_link_libraries(_espressopp ${Boost_LIBRARIES} ${PYTHON_LIBRARIES} ${MPI_LIBRARIES} ${FFTW3_LIBRARIES} ${HDF5_LIBRARIES} ${VAMPIRTRACE_LIBRARIES})
+endif()
+
```
junghans: With empty HDF5_LIBRARIES no need for a conditional.

CMakeLists.txt
fedepad: Well if it's not found somehow has to be fetched and built right?

src/CMakeLists.txt
fedepad: That's very true, I realized later but still didn't change...

junghans:
@MrTheodor : A more general question is, hdf5 vs h5md format, why are we supporting both?

CMakeLists.txt
```diff
@@ -160,6 +161,76 @@ else()
...
+########################################################################
+#HDF5 settings
+########################################################################
+
+
+if(ENABLE_HDF5)
+    find_package(HDF5)
+    if(HDF5_FOUND)
+        include_directories(${HDF5_INCLUDE_DIR})
+        set(_hdf5_libs hdf5 )
```
junghans: Isn't that support to be `HDF5_LIBRARIES` instead of `_hdf5_libs`?

MrTheodor:
@junghans huh I don't know, H5MD is a file format that uses HDF5 as a storage but in principle it could be implemented on any storage, not only HDF5.

MrTheodor:
@fedepad would be it possible to clean it a bit by removing all commented code..

fedepad:
@MrTheodor should be a little bit more clean!

fedepad:
Again, the the testsuite will fail for the line
from espressopp.io.HDF5File import *
in src/io/__init__.py

junghans:
Is your hdf5 format the same spec as @MrTheodor's h5md format?

junghans:
I guess you could add `libhdf5-dev` to the package to be installed in the `.travis.yml` file to fix the build.

src/io/__init__.py
```diff
@@ -25,3 +28,4 @@
 from espressopp.io.DumpXYZ import *
 from espressopp.io.DumpGRO import *
 from espressopp.io.DumpGROAdress import *
+from espressopp.io.HDF5File import *
```
junghans: Couldn't you wrap this line into a try block?

fedepad: that's good!

fedepad:
@junghans @MrTheodor  I should look into the H5MD format. Probably it's not strictly the same as I'm saving only some info in a simple dataset, so probably the file structure it's different at the moment since H5MD should be a little bit more elaborate as to store info (nested groups, etc.).
As far as I understood from @MrTheodor DumpH5MD requires h5py. Do I remember correctly @MrTheodor ? This version doesn't require anything outside the parallel (or serial, but then we need to modify something) version of the plain HDF5 library, no extra python packages or others.

junghans:
I would really prefer using the h5md spec over any custom file format. h5md is supported by VMD, VOTCA and a couple of other packages. Please let's not re-invent the wheel, again!

MrTheodor:
There is some implementation of h5md in C: https://github.com/pdebuyl/ch5md 
Maybe you could integrated that with your code of parallel writing.

I am wondering what will be the performance of those two solutions.

junghans:
Ping!

fedepad:
@junghans Thanks for reminding me, I appreciate it; I will follow up.
Thanks again, right now I'm on a tight schedule at LLNL.

fedepad:
@MrTheodor I will do that! Thanks for pointing to the C implementation of h5md!

MrTheodor:
@fedepad maybe this will be also useful: https://github.com/espressomd/espresso/pull/745

fedepad:
@MrTheodor thanks! Yes, it looks they use h5xx, which I tried to use it many (probably a little bit over 6 months ago) months ago and I had some issues. I have the new version so I'm willing to give it another try, but basically it's a wrapper and if I remember it uses the C HDF5 interface, since the C++ interface probably is not super robust as the C (I played at the beginning with the C++ HDF5 interface so that everything could be C++ but I was getting issues with the same code which were not present when using the C interface). I should check the parallel io support. Thanks for pointing to espressomd for that.

govarguz:
ping @fedepad

fedepad:
I'm turning this into H5MD format as suggested by other people, because they see more use. This version is not expected to rely on h5py for example, just on the HDF5 C library, just as this prototype. Although I really don't like to have doubles and it seems that in general Jakub started way earlier with H5MD I still think would be good to compare how the two solutions work before landing one (extra layer of h5py worth it for simplicity or might hurt a bit times?).

As I've been emailing @MrTheodor since I've been back a couple of times I'm waiting to test his version #52  since the HPC cluster for benchmark is having FS big problems for now about 12 days, and are still not solved. Since we test writing files that's important and relevant, as we can cause a lot of unexpected I/O errors in GPFS and we wouldn't get reliable measurements and probably even generate corrupted files. 

Just wanted to point out for the few interested in some technical details, that the main difference between one of H5MD implementations (e.g. ch5md, DumpH5MD) and this prototype with HDF5 that dumps the particles data structure is that this prototype write a file contiguously while those others write chunked datasets, which has many advantages and its performance depends mainly on the chunk size: this also has another side and it means that with a "wrong" choice of the chunk size performance might be even worst.

fedepad:
Update: Currently writing H5MD files by leveraging ch5md (included headers, etc. w/ some modifications). Right now the writing patterns available are N-to-1 (one shared file) and N-to-N (a file per rank) by turning on and off a switch at the python level.
Running benchmarks for writing at the moment.
Code is setup at the moment in a way that allow me to run benchmarks to check if a varying number of writers/aggregators at the MPI layer with respect to the number of MPI ranks used by the application will affect the performance at a little larger scales (e.g. application uses 1024 MPI ranks, trying to write with 512 for example). This uses MPI hints and relies on the MPI implementation to pass this info to lower layers.
Currently this version doesn't rely on h5py (the python "version" of HDF5) or any other external python packages, just on plain parallel HDF5 package, C/C++ and the usual E++ stack (e.g. Boost.python).

junghans:
Ping @fedepad

fedepad:
@junghans From what I have read the team already decided a while ago for @jkrajniak version:
https://arxiv.org/pdf/1806.10841.pdf
His version is the closest I can imagine to the description given in the 7th paragraph. And December 2018 it is the date when it was made available so I expect this was already the status many months back.
There is also #187 still open, so I am a bit confused about that.
I had a version here:
https://github.com/fedepad/espressopp/tree/io-h5md/src/io
which writes h5md files by using the ch5md library (included so no extra dependencies). 
The main files are:
- H5MDWriter.cpp
- H5MDWriter.hpp
- H5MDWriter.py
I do not remember how much performance and other kind of tests I did on this at that time.
To me it looks that @jkrajniak already had a working software before I even started working on this. Based on this and the description in 7th I think this is obsolete in comparison.

fedepad:
Obsolete given https://github.com/espressopp/espressopp/pull/356
In my opinion it should be closed.

XzzX:
@junghans Why didn't you merge it in the last 5 years? Are there serious problems?

I will support and maintain my version.

junghans:
Merge conflicts....

XzzX:
So if no one is willing to fix the conflicts it cannot be merged. Should we close it?

junghans:
@stuehn what do you want to do with this code?

XzzX:
HDF5 support added with #356

Issue #34, opened 2015-11-27
Issues with different combinations of boost, gcc, apple clang, etc


niktre:
Hi! I wasn't sure whether to re-open #8 or start a new issue.. 
I have a pleasure of using a MacBook and therefore have additional fun compiling ESPResSo++, but it is interesting NOT only for Mac users:
Boost: external 1.59.0
MPI4PY: internal
c / c++: from Apple (clang 7.0.0)
c / c++ flags: -std=c99 / -std=c++03 

The error I get when compiling:
Linking CXX shared library ../_espressopp.so
Undefined symbols for architecture x86_64:
  "boost::system::system_category()", referenced from:
      __GLOBAL__sub_I_bindings.cpp in bindings.cpp.o
      __GLOBAL__sub_I_DumpGRO.cpp in DumpGRO.cpp.o
      __GLOBAL__sub_I_DumpGROAdress.cpp in DumpGROAdress.cpp.o
      __GLOBAL__sub_I_DumpXYZ.cpp in DumpXYZ.cpp.o
  "boost::system::generic_category()", referenced from:
      __GLOBAL__sub_I_bindings.cpp in bindings.cpp.o
      __GLOBAL__sub_I_DumpGRO.cpp in DumpGRO.cpp.o
      __GLOBAL__sub_I_DumpGROAdress.cpp in DumpGROAdress.cpp.o
      __GLOBAL__sub_I_DumpXYZ.cpp in DumpXYZ.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [_espressopp.so] Error 1
make[1]: *** [src/CMakeFiles/_espressopp.dir/all] Error 2
make: *** [all] Error 2

If I switch to internal Boost, everything is fine!

junghans:
Does that help? (`git pull https://github.com/junghans/espressopp.git master`)

junghans:
Somebody needs to look into that: <https://docs.travis-ci.com/user/osx-ci-environment/>

niktre:
Sorry Christoph, I will be able to check if it works only on Tuesday..

niktre:
[...]

PR #62, opened 2016-02-21
.travis.yml: test against different gcc version

junghans:


junghans:
@niktre : that is half of #34.

Issue #39, opened 2015-12-11
Clean up dirty code base


MrTheodor:
There is a lot of abandon commented code in the code base e.g. MDIntegrator or VerletList has whole blocks of it; some leftovers after print debugging which are not necessary correct if you uncomment them.
Therefore it is quite painful for newcomers to read it and to spot any bug.

junghans:
:+1: Running uncrustify on the code won't hurt as well.

govarguz:
@fedepad, status update...pls!

PR #63, opened 2016-02-27
Example of running uncrustify on the src/storage 

MrTheodor:
Here is an example of changes that uncrustify can do on the code in src/storage. Looks pretty nice, IMHO. Uncrustify only reformat code so there should be no issue with any missing parts of code.  

There is a comprehensive code style for C++: https://google.github.io/styleguide/cppguide.html
with cpplint that does check https://github.com/google/styleguide/tree/gh-pages/cpplint 
Still this uncrustify does not solve all problems pointed out by cpplint but the code looks
cleaner.
Perhaps the solution would be to use cpplint as a git hook, unfortunately it can be only locally so
it is rather a case agreement. At least it will prevent from committing 
the files without a proper code style.

cc: @stuehn @govarguz @junghans 
issue #39

### DIFF
--- a/src/storage/CellGrid.cpp
+++ b/src/storage/CellGrid.cpp
@@ -1,24 +1,24 @@
 /*
-  Copyright (C) 2012,2013
+   Copyright (C) 2012,2013
       Max Planck Institute for Polymer Research
-  Copyright (C) 2008,2009,2010,2011
+   Copyright (C) 2008,2009,2010,2011
       Max-Planck-Institute for Polymer Research & Fraunhofer SCAI
-  
-  This file is part of ESPResSo++.
-  
-  ESPResSo++ is free software: you can redistribute it and/or modify
-  it under the terms of the GNU General Public License as published by
-  the Free Software Foundation, either version 3 of the License, or
-  (at your option) any later version.
-  
-  ESPResSo++ is distributed in the hope that it will be useful,
-  but WITHOUT ANY WARRANTY; without even the implied warranty of
-  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-  GNU General Public License for more details.
-  
-  You should have received a copy of the GNU General Public License
-  along with this program.  If not, see <http://www.gnu.org/licenses/>. 
-*/
+
+   This file is part of ESPResSo++.
+
+   ESPResSo++ is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   ESPResSo++ is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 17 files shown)


junghans:
Looks good to me! We could even make a travis job to run uncrustify and fail if there is a diff.

MrTheodor:
:+1: for travis, I will have a look on that

junghans:
I could add an cmake `uncrustify` target, which you can run in the `after_sucess` block if `[[ ${UNCRUSTIFY} = "yes ]]` and then use `git diff --exit-code`.

fedepad:
@junghans @MrTheodor What happened to the checks that have not been completed? codecov/project says "waiting for status to be reported". Was it down at the time was run?
Or there's something I'm missing here.

fedepad:
I'm looking also into this and include a cmake target for it as @junghans mentioned.
Since this involves a given configuration of uncrustify a.k.a. a codying style, I think people should agree on a style (or a set of options for uncrustify config file).
Of course, if we fail when a diff is detected and to prevent that, will mean that contributors should run it on their local machines before pushing/opening a PR, or something close to it.
I agree 100% this is good practice.

Just asking if people are on board with it.

MrTheodor:
Huh, I haven't look into it for a while. 
Uncrustify should prepare the current codebase to be compatible with any syntax check up like cpplint (https://pypi.python.org/pypi/cpplint)
If the codebase is clean then the travis can handle checking it with every push. I think that the best way, otherwise there would be always a reason to not follow any code style which is currently the case.
The same should be done with Python code, there are linters like pep8, pyflakes or pylint but I haven't seen any "python-uncrustify" code.

Modern IDE like CLion can point out and reformat code according to defined style. With a bit of effort, vim can also very easily show result from local run of cpplint (and highlight the errors during the coding).

fedepad:
@MrTheodor I agree 100%! Yes, even the old Eclipse CDT should be able to do it, as I used for example google style for some cpp and they have a config file to load in eclipse:

https://github.com/google/styleguide

As you can see they have also a config for python style for vim.

I know of a python project which enforces through Travis PEP8 compliance of the submissions.
They should have a script (if I remember correctly) that one can even run locally on the python files
before submission. I will have a look on those...

niktre:
which milestone would you like to attach here?

MrTheodor:
I think that having it already in 2.0.0 would be a good idea.

fedepad:
@niktre I agree with @MrTheodor .
I'm playing around with it to see if it makes sense to have it as git hooks as either pre-commit or pre-receive on the servers or just having the cmake target, etc. as @junghans said.
Also then we need some discussion about which way to go.

fedepad:
I will assign the 2.0.0 milestone if everyone is ok with it.

MrTheodor:
@fedepad problem with local hook is that it is local ;) not that we cannot agree on something but having watchdog on travis to prevent commits that does not pass ```cpplint``` seem to me a better idea. 
Of course one can have local hook that will check commit.
Github supports server side hooks only in the enterprise version.

niktre:
Do it!

Von meinem iPhone gesendet

PR #73, opened 2016-04-18
Minor fixes to python-level tools and documentation

acfogarty:


src/tools/pdb.py
```diff
@@ -177,15 +185,18 @@ def psfwrite(filename, system, maxdist=None, molsize=4, typenames=None):
       ypos   = particle.pos[1]
       zpos   = particle.pos[2]
       type   = particle.type
+      q      = particle.q
+      mass   = particle.mass
       if typenames:
 	  name=typenames[type]
-      st = "%8d T%03d %4d UNX  %2s   %2s                    \n" % (pid+addToPid, type, mol, name, name)
+      st = "%8d T%5d    UNX  %-4s %-4s  %9.6f%14.4f\n" % (pid+addToPid, mol, name, name, q, mass)
```
junghans: I got very confused by looking at this line not realizing that it is about the psf format, I guess the psf functionality should be move into its own file.

junghans:
Please update the copyright headers as well.

src/tools/convert/gromacs.py
```diff
@@ -825,6 +825,7 @@ def storeDihedrals(f, types, dihedraltypes, dihedraltypeparams, dihedrals, num_a
     return dihedrals
     
 def storeImpropers(f, types, impropertypes, impropertypeparams, impropers, num_atoms_molecule, num_molecule_copies, molstartindex, atomtype_wildcard):
+    print '#Warning! This parser of the improper angles section of gromacs-format forcefield files is for the Amber forcefield only. Other forcefields may have a different atom ordering.'
```
junghans: I would add a comment to look `gromacs.py` for details and/or open a bug on github to implement addition functionality.

acfogarty: I'll go with a comment pointing the user to the details in gromacs.py, because actually there are a lot of different protein forcefields with various different ways to treat impropers, so if we open bugs for them all we'll have a long list of bugs, without any user ever actually having requested the functionality.

src/tools/pdb.py
acfogarty: Ok, agreed.

niktre:
Looks nice!

PR #74, opened 2016-04-20
Fix a bug in FENE interaction

jsmrek:


src/interaction/FENE.hpp
```diff
@@ -97,13 +104,11 @@ namespace espressopp {
 
         real ffactor;
         
-        if(r0 == 0) {
+        if(r0 != 0) {
```
junghans: Uh, ugly! A comparison `r>A_SMALL_NUMBER` would be better!

junghans:
And please update copyright in `src/interaction/FENE.hpp`.

govarguz:
Besides the comments already made by @junghans and if @stuehn has no additional ones. It looks OK to me

MrTheodor:
Please check if the documentation is correct. 
Would it be possible for you to write some sort of test to check the correctness (in testsuiet/) ?

jsmrek:
@junghans Ok, I'll do the changes, what small number would you suggest?
@MrTheodor I'm not quite sure, what documentation you are referring to. 
It makes more sense to me if someone else would check it against her\his calculations. Of course, I checked myself the correctness for a bunch of parameters by plotting forces and energies and comparing with analytical results.

MrTheodor:
@jsmrek I mean the one in .hpp file that goes to developer doc and the one in doc/ug/;

niktre:
@MrTheodor Do not want that you get me wrong, but since when do we actually care about developer doc? 

@jsmrek use `real threshold_ = 1e-30`. I think it is enough.

For user documentation that @MrTheodor mentions, it would be great if you provide some description of the FENE potential in the python file. As a quick reference, look at the `LennardJones93Wall.py` (lines 23 till 43). There you'll find the formula (written in `.. math:` environment), and a couple of usage examples. If you are not sure about your documentation in python-file (whether it makes sense, for example :)), check your email. 

Do not work at the weekend! Remember, you have a couple of girls to take care about!!

MrTheodor:
@niktre I don't know, but it exists so why not to keep it updated or remove it; IMHO
@jsmrek there is a definition of very small number in `src/include/esconfig.hpp` `ROUND_ERROR_PREC`.

niktre:
It was a long way from a simple bug to an acceptable documentation! 👍

PR #78, opened 2016-05-22
new LangevinThermostatOnGroup that thermalize only particles in given

MrTheodor:
- extension to LangevinThermostat to work on the ParticleGroup instead of all particles.


govarguz:
@MrTheodor, the extension looks fine. However, I would like to see a test or an example of use for the <LangevinThermostatOnGroup>.... before commenting further!

govarguz:
The unit test looks fine, as well. I would like to test some performance, still. Maybe it could be done automatically with every PR, but I did not manage to further look for it. Could you add some performance runs with old and new versions on same nr. of particles w and wo the new feature?

MrTheodor:
Simple benchmark:
Standard LV: 5.68 +/- 0.03 s
Group LV: 5.75 +/- 0.01 s

5000 LJ particles, on 20 cores;

Files:
https://dl.dropboxusercontent.com/u/525038/langevinongroup.zip

src/integrator/LangevinThermostatOnGroup.py
```diff
@@ -0,0 +1,52 @@
...
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+r"""
+***************************************************
+**espressopp.integrator.LangevinThermostatOnGroup**
+***************************************************
+
+Thermalize particles in the ParticleGroup only.
+
```
acfogarty: Could we also include a few lines of example python script here? Showing creation of particle group and then adding thermostat extension to integrator.

MrTheodor: Done

niktre:
I like it

PR #82, opened 2016-07-08
default constructor now returns 0 energy and force

acfogarty:
in ReactionFieldGeneralized and CoulombTruncated.
This is necessary because potentials' default
constructors are used to fill in empty entries
in the 2D potential types arrays.

### DIFF
--- a/src/interaction/CoulombTruncated.hpp
+++ b/src/interaction/CoulombTruncated.hpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2012,2013,2015
+  Copyright (C) 2012,2013,2015,2016
       Max Planck Institute for Polymer Research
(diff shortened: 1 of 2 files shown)


kkreis:
Just to be consistent: Should we also change the default of the prefactor on the python level to 0 then?

acfogarty:
About changing the default on the python level: I would say no. The default on the c++ level is for filling in the potential array. The default on the python level is for the user who doesn't want to choose parameters, but just wants to take some standard default values. So they're two completely separate things.

See for instance how it works in the other potentials like LennardJones, with 1 in the .py and 0 in the .cpp

kkreis:
I was just wondering, because setting the prefactor to 1.0 also doesn't make much sense physically. But you are right, this is a different question. So, let's keep it as is.

PR #83, opened 2016-07-15
Several fixes and improvements on AdResS and related stuff

kkreis:
* Added flexible AdResS feature (see Aoife/my JCTC paper)
* Added analysis routine to compute radial density profile in AdResS simulations
* Improved the RDF analysis rountine for AdResS simulations
* Added exclusions in the Langevin Thermostat
* Finally added spherical options for H-AdResS simulations
* Furthermore, in all touched files old comments and old commented code was cleaned


src/analysis/RDFatomistic.cpp
```diff
@@ -40,41 +40,34 @@ using namespace std;
 
 namespace espressopp {
   namespace analysis {
-    
-    // TODO currently works correctly for Lx = Ly = Lz
+
+    // TODO currently works correctly if L_y is shortest side length (it's fine if L_x or L_z are equal)
```
acfogarty: You mean if L_x or L_z are equal to L_y, right? Not if they're equal to each other? Also, we should explictly mention this limitation in the documentation in .py

src/analysis/RDFatomistic.cpp
```diff
@@ -86,114 +79,128 @@ namespace espressopp {
...
+                real resolutionP2 = config[j].resolution;
+
+                if (molP1 != molP2){
+
+                    if( ( (typeP1 == target1) && (typeP2 == target2) ) || ( (typeP1 == target2) && (typeP2 == target1) ) ){
+
+                        num_pairs += 1;
+
+                        if( resolutionP2 == 1.0 ){
+                          Real3D distVector = coordP1 - coordP2;
```
acfogarty: You don't ever need to use PBC in this class?

src/integrator/LangevinThermostat.cpp
```diff
@@ -128,11 +129,16 @@ namespace espressopp {
...
 
       System& system = getSystemRef();
-      
+
       CellList cells = system.storage->getRealCells();
 
       for(CellListIterator cit(cells); !cit.isDone(); ++cit) {
-        frictionThermo(*cit);
+
+        if(exclusions.count((*cit).id()) == 0)
```
acfogarty: We're adding extra overhead here for a situation that rarely occurs (I guess 99% of the time people will use thermostat with no exclusions), but I guess if length(exclusions)==0 then the call to count has almost no cost...

src/integrator/LangevinThermostat.py
```diff
@@ -1,55 +1,77 @@
...
+>>> # add extensions to a previously defined integrator
 
 .. function:: espressopp.integrator.LangevinThermostat(system)
 
-		:param system: 
-		:type system: 
+        :param system: system object
+        :type system: shared_ptr<System>
+
+.. function:: espressopp.integrator.LangevinThermostat.addExclusions(pidlist)
```
acfogarty: Need to say if in AdResS both AT and CG pids need to be in pidlist

src/analysis/RDFatomistic.py
```diff
@@ -1,46 +1,60 @@
...
 
 r"""
 ************************************
 **espressopp.analysis.RDFatomistic**
 ************************************
 
+Class to compute radial distribution functions in adaptive resolution simulations in subregions of the box.
 
-.. function:: espressopp.analysis.RDFatomistic(system, type1, type2, _span)
+On the one hand, the RDF can be calculated in a cuboid region in the center of the box (periodic in y,z, limited in x). In this case, particle pairs are considered for which at least one of them is in the defined cuboid region (spanbase = True). This can be useful when the high resolution region has a slab geometry. No further normalization should be required. On the other hand, the routine can also calculate unnormalized RDFs using particle pairs with both particles being in the high resolution region. This can be useful when atomistic region has complicated or spherical geometries. In any case, only atoms belonging to different molecules (i.e. different coarse-grained entities) are considered.
```
acfogarty: How about: 

"OPTION 1 (SPANBASE=TRUE) : the RDF can be calculated in a cuboid region in the center of the box (periodic in y,z, limited in x). In this case, particle pairs are considered for which at least one of them is in the defined cuboid region. This can be useful when the high resolution region has a slab geometry. No further normalization should be required. 

OPTION 2 (SPANBASE = FALSE):  the routine can also calculate unnormalized RDFs using particle pairs with both particles being in the high resolution region. . This can be useful when atomistic region has complicated or spherical geometries. 

In any case, only PAIRS OF ATOMISTIC PARTICLES belonging to TWO different COARSE-GRAINED PARTICLES are considered." 

Did I understand correctly?

src/analysis/RDFatomistic.py
```diff
@@ -1,46 +1,60 @@
...
+
+>>> rdf_0_1.compute(100)
+>>> # computes the rdf using 100 bins over a distance corresponding to L_y / 2.0
+
+.. function:: espressopp.analysis.RDFatomistic(system, type1, type2, span, spanbased)
+
+                :param system: system object
+                :param type1: type of atom 1
+                :param type2: type of atom 2
+                :param span: (default: 1.0) radius of the cuboid region used for rdf calculation (if spanbased == True)
```
acfogarty: Because span appears before spanbased in the list of arguments, it looks like span is always used if spanbased is used. Need to explicitly say here that span is not used if spanbased == False. Also need to say explicitly somewhere in documentation that if spanbased == False, the decision about what particles to include is based on their value of lambda_adr, i.e. as determined elsewhere in the code. Otherwise it looks like the value of span is being used to define the high resolution region when spanbased == False.

acfogarty: Instead of radius, let's say distance +/- centre of box in x-direction.

src/analysis/RDFatomistic.py
```diff
@@ -1,46 +1,60 @@
...
-		:param _span: 
-		:type system: 
-		:type type1: 
-		:type type2: 
-		:type _span: 
+Examples:
+
+>>> rdf_0_1 = espressopp.analysis.RDFatomistic(system = system, type1 = 0, type2 = 1, span = 1.5, spanbased = True)
+>>> # creates the class for calculating the RDF between atomistic particles of type 1 and 0 between different molecules,
+>>> # within plus/minus 1.5 from the center of the box in x-direction
```
acfogarty: It's only necessary for one of them to be within plus/minus 1.5 from the center of the box in x-direction, right? This sounds like both of them should be.

src/integrator/Adress.cpp
```diff
@@ -413,95 +368,141 @@ namespace espressopp {
...
     // AdResS Weighting function
     real Adress::weight(real distanceSqr){
         if (dex2 > distanceSqr) return 1.0;
         else if (dexdhy2 < distanceSqr) return 0.0;
         else {
             real argument = sqrt(distanceSqr) - dex;
-            return 1.0-(30.0/(pow(dhy, 5.0)))*(1.0/5.0*pow(argument, 5.0)-dhy/2.0*pow(argument, 4.0)+1.0/3.0*pow(argument, 3.0)*dhy*dhy);
-            //return pow(cos(pidhy2 * argument),2.0); // for cosine squared weighting function
+            //return 1.0-(30.0/(pow(dhy, 5.0)))*(1.0/5.0*pow(argument, 5.0)-dhy/2.0*pow(argument, 4.0)+1.0/3.0*pow(argument, 3.0)*dhy*dhy);
+            return pow(cos(pidhy2 * argument),2.0); // for cosine squared weighting function
```
acfogarty: Can you make the same change in VerletListAdressInteractionTemplate.hpp? It occurs twice, just search for cos, the commented lines are already there.

testsuite/CMakeLists.txt
```diff
@@ -5,4 +5,8 @@ add_subdirectory(tabulated)
 add_subdirectory(system_test)
 add_subdirectory(minimize_energy)
 add_subdirectory(Array4DTest)
-add_subdirectory(fixed_list_types_test)
\ No newline at end of file
+add_subdirectory(fixed_list_types_test)
+add_subdirectory(RDFatomistic)
+add_subdirectory(AdressDensity)
+add_subdirectory(TDforce)
+add_subdirectory(FreeEnergyCompensation)
```
junghans: I think all these tests should go into an `AdResS` subfolder.

kkreis: Possibly a good idea. How's that being done properly? Do I need to add just another CMakeLists.txt like this one in this subfolder specifying the different folders like here?

junghans: I would move the 4 tests to `AdResS/TDforce` etc. and add a `CMakeLists.txt` in the `AdResS` folder and do `add_subdirectory(AdResS)` here.

testsuite/AdResS/FreeEnergyCompensation/test_FreeEnergyCompensation.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # run ten steps and compute energy
+        integrator.run(10)
+        energy = fec.computeCompEnergy()
+
+        # x coordinates of particles after integration
+        after = [self.system.storage.getParticle(i).pos[0] for i in range(1,6)]
+
+        # run checks
+        self.assertAlmostEqual(before[0], after[0], places=5)
```
acfogarty: I don't understand what's happening here. Why are some of the particles moving and others not?

kkreis: It checks really only the FEC module without any potentials and stuff in the system. I wanted to keep it as small as possible only checking what it says, i.e. the FEC. Hence, only particles in the hybrid region which feel the FEC move. This is the case only for one of the particles.

kkreis: I can do that. Would you suggest to explain the "ideas" behind my unit tests in each of them in short comments?

kkreis: Done. Added some short explanations in the non-obvious tests.

kkreis:
@acfogarty I now successfully merged also your latest commit on thermodynamic integration. Please have a final look. Thanks!

acfogarty:
@kkreis It looks good. I'm just going to run some of the AdResS unit tests on an old version of the code (since in the ideal case we would already have had those unit tests, and they would have told us your pull request wasn't changing anything in existing code, rather than the other way round as it is now!)

govarguz:
@kkreis you need to comment parts of the code that would need improvements in terms of performance. This will help to identify where the code is overloaded. I would suggest you use some tag "need efficiency improvements"

kkreis:
@govarguz Well, in the files I changed there is nothing which would need any particular relevant performance improvements, I think. Esp++ in general or its AdResS implementation as a whole is not optimal - but this has nothing to do with this pull request and can of course not be assigned to any particular lines of code, especially not in the files touched in this pull request.

govarguz:
@kkreis OK! then we assume performance related to this PR is untouched, right? Did you checked that?

kkreis:
@govarguz I did not run any detailed tests, but this pull request changes essentially no old stuff at all. Old code will run in exactly the same way as before. It only adds new functionality that was not present before. Hence, nothing to compare against.

govarguz:
@kkreis, if under your Opinion there is no room for improvement in the new functionalities e.g. "flexible AdResS" then no comments will be needed. However, what about comparing the flexible AdResS approach with a moving big sphere? Is this part of the "roadmap" or was done?

kkreis:
@govarguz Of course, one could do a quantitative study of the speed based on the number of AdResS regions in the flexible AdResS approach for various systems and system parameters. However, the actual corresponding increase/decrease in speed depends as usual on a large number of system parameters. A meaningful detailed study is way beyond the scope of this pull request... This could be a whole paper. Furthermore, this would anyhow need to be tried out and optimized for every new system... Additionally, this issue has been addressed in Aoife's/my paper (the reference is given) already. Also note that this is mainly a proof of principle work. This kind of stuff only becomes relevant for really large systems which you really do not want to simulate with Esp++ anyhow ;) So, of course, the implementation could be optimized and there is, in principle, a whole lot of room for improvement, but as I said before, this would, I think, at the end of the day mean a complete remodeling of whole AdResS in Esp++ or rather Esp++ itself.

All that being said, I am happy to add an additional comment for the user in the description in the python file.

govarguz:
@kkreis of course, it is a "Proof of principle". However, this PR aims to add a functionality that when a new user want to employ might get surprised! Therefore, your comments on performance of this implementation are needed. Moreover, a simple comparison for the same parameters between flexible AdResS approach to a moving big sphere in AdResS will add much more insight to the user

kkreis:
@govarguz I added some comments that clarify what's going on. Now no user will be surprised anymore. 

A simple comparison on some random system which one would in practice never treat with this method anyhow could be even highly misleading as the behavior will most certainly be very different for other systems. To repeat myself: A meaningful study on truly relevant systems (huge folding proteins or deforming membranes, for example) is way beyond the scope of this simple pull request. This would be a whole project/paper. Once again: The main aim of this pull request is to only provide the functionality that was used in Aoife's/my paper and to provide a proof of principle.

testsuite/AdResS/HAdResS/test_HAdResS.py
```diff
@@ -0,0 +1,164 @@
...
+        # generate a verlet list
+        vl = espressopp.VerletListAdress(self.system, cutoff=1.5, adrcut=1.5,
+                                dEx=2.0, dHy=1.0, adrCenter=[5.0, 5.0, 5.0], sphereAdr=False)
+
+        # add interaction
+        interNB = espressopp.interaction.VerletListHadressLennardJones2(vl, ftpl)
+        potWCA1  = espressopp.interaction.LennardJones(epsilon=1.0, sigma=1.0, shift='auto', cutoff=1.4)
+        potWCA2 = espressopp.interaction.LennardJones(epsilon=0.0, sigma=1.0, shift='auto', cutoff=1.4)
+        interNB.setPotentialAT(type1=0, type2=0, potential=potWCA1) # AT
+        interNB.setPotentialCG(type1=0, type2=0, potential=potWCA2) # CG
```
acfogarty: It's `type1=1, type2=1` (not that makes a difference in this particular test, but I noticed when I tried to copy and extend the test)

kkreis: Very true, thanks. I fixed it.

acfogarty:
It looks like this is ready to merge, unless you have any more comments @govarguz ? It looks like @kkreis addressed all your concerns about performance.

govarguz:
@kkreis, to be clear, I am not asking for "(huge folding proteins or deforming membranes, for example) is way beyond the scope of this simple pull request" or whatsoever. Rather to use exactly the same "proof of principle system" of the paper you are citing and compare in terms of performance if the flexible-AdResS functionality with a moving spherical-AdResS setup. In other words, to build a proof of performance of the "system". This will be useful from the perspective of the user and will serve as a guide. If you add a functionality you may expect that someone will use it, right? However, if this "proof of principle system" is misleading in terms of performance and the huge project you mentioned is needed. @acfogarty  I would have no more comments!

kkreis:
@govarguz As I mentioned before, the performance issue is a bit subtle for this case, as the system in the paper is only a "proof of concept/functionally" but not really suited for a "proof of performance". However, Aoife and me are planning to add the system to the examples in the near future, where we may nevertheless add a similar test as you suggested. In any case, the new comment in the .py documentation should clarify the situation for the user. Hence, thanks for your feedback.
@acfogarty Feel free to merge.

PR #89, opened 2016-08-30
Fixed Minimizeenergy and implementation of variable Gamma step

hidekb:
Fixed MinimizeEnergy of measurement  of particles movement distance.
Implementation of the variable Gamma step in steepestDescentStep().

### DIFF
--- a/src/integrator/MinimizeEnergy.cpp
+++ b/src/integrator/MinimizeEnergy.cpp
@@ -22,171 +22,193 @@
 #include "MinimizeEnergy.hpp"
 
 namespace espressopp {
-namespace integrator {
+    namespace integrator {
+	
+	using namespace interaction;
+	using namespace iterator;
+	using namespace esutil;
+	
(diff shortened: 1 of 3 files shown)


govarguz:
It looks neat and more than adding variable Gamma. It is much faster than the former version since particle movement has been initialized and goes along to Fmax. However you need to update the copyright!

junghans:
Btw, one can get a diff without whitespace changes by adding `?w=1` to diff url, e.g.:
<https://github.com/espressopp/espressopp/pull/89/files?w=1>

MrTheodor:
@junghans thanks for tip, better than nothing... but this is rather a case of not having defined code style

Although code was already pushed, the text in docs does not reflect the variable gamma flag anymore. And it would be good to have unit test that will cover the flag and perhaps the error with max_DP.

hidekb:
Thanks you for the suggestion. I will add the explanation of using a variable gamma flag.

PR #91, opened 2016-09-08
bugfix for Cosine potential

acfogarty:
Fix supplied by Hsiao-Ping Hsu. In the old code, calculation of energy was inconsistent with definition of angle and with calculation of force. Also, theta0 was ignored in the force calculation. This is all corrected in the new code.

### DIFF
--- a/src/interaction/Cosine.hpp
+++ b/src/interaction/Cosine.hpp
@@ -72,6 +72,8 @@ namespace espressopp {
 			    const Real3D& dist12,
 			    const Real3D& dist32) const {
 
+      	const real SMALL_EPSILON = 1.0E-9;
+
         real dist12_sqr = dist12 * dist12;
         real dist32_sqr = dist32 * dist32;
         real dist12_magn = sqrt(dist12_sqr);
@@ -90,12 +92,13 @@ namespace espressopp {
         a12 = -1.0 / (dist12_magn * dist32_magn);
         a22 =  cos_theta / dist32_sqr;
 
-        if (sin_theta !=0 ) {
+        if (sin_theta > SMALL_EPSILON) { //sin_theta is always positive
           force12 = (Kcos_theta0-Ksin_theta0*cos_theta/sin_theta)*(a11 * dist12 + a12 * dist32);
           force32 = (Kcos_theta0-Ksin_theta0*cos_theta/sin_theta)*(a22 * dist32 + a12 * dist12);}
         else {
-          force12  = 0.0*dist12;
-          force32  = 0.0*dist32;}
+          force12  = Real3D(0.0,0.0,0.0);
+          force32  = Real3D(0.0,0.0,0.0);
+        }
 
         
         return true;
@@ -104,14 +107,6 @@ namespace espressopp {
       // used for generating tabular angular potential
       real _computeForceRaw(real theta) const {
 
-        /*
-        real cos_theta = cos(theta);
-        if(cos_theta < -1.0) cos_theta = -1.0;
-        else if(cos_theta >  1.0) cos_theta =  1.0;
-
-        real sin_theta = sqrt(1.0 - cos_theta * cos_theta);
-        return K * sin_theta;
-        */
         return K;
         
       }
--- a/src/interaction/Cosine.hpp
+++ b/src/interaction/Cosine.hpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2012,2013
+  Copyright (C) 2012,2013,2016
       Max Planck Institute for Polymer Research
   Copyright (C) 2008,2009,2010,2011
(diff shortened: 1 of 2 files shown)


MrTheodor:
Could you add the unit test?

8 wrz 2016 10:09 "acfogarty" <notifications@github.com> napisał(a):

acfogarty:
@MrTheodor Because this is a bug fix I'd like to get it accepted without writing the unit test right now. If we make a rule that people have to write unit tests and fully document a class before fixing any bugs in it, people will just delay fixing bugs...

However, actually I was just about to open an issue about unit tests in general, because this morning I was writing some for some new code of my own, and thinking about this. Maybe you can give me some ideas over in Issues?

govarguz:
OK, This bug fix was announced to appear a while ago. However I would still change the part of the code where force12 = 0.0*dist12 and force32  =0.0*dist32, shouldn't we stick to "=0.0"? 
On the other hand, it seems to be that the editor changed smth because Cosine.hpp is a completely new file while for Cosine.py only the lines changed appear.

acfogarty:
@govarguz It 's force12 = 0.0*dist12 and not just force12=0.0 because the forces are Real3D, not real. Or one could also create a new Real3D containing 3 zeros, but this isn't necessary.

govarguz:
@acfogarty sure! I meant Real3D abcdf(0.0, 0.0, 0.0); to avoid multiplying it so often. However it seems to be this condition won't be fulfill so often... this maybe an input of Hsiao-Ping

acfogarty:
@govarguz  Ah okay, got it. Yes, I think this condition rarely or never occurs, since in the old version that if-check didn't even exist, yet people were still using the code. Still, Hsiao-Ping put in the check just in case, certainly a good idea.

junghans:
Ignore white spaces in diff: https://github.com/espressopp/espressopp/pull/91/files?w=1

src/interaction/Cosine.hpp
```diff
@@ -1,107 +1,124 @@
...
+        if(cos_theta < -1.0) cos_theta = -1.0;
+        if(cos_theta >  1.0) cos_theta =  1.0;
+
+        real sin_theta = sqrt(1.0 - cos_theta*cos_theta);
+ 
+        a11 = cos_theta / dist12_sqr;
+        a12 = -1.0 / (dist12_magn * dist32_magn);
+        a22 =  cos_theta / dist32_sqr;
+
+        if (sin_theta !=0 ) {
```
junghans: It might be better to compare against small number, e.g.:
```
if (sin_theta < SMALL_EPSILON) {
```
(see https://github.com/espressopp/espressopp/blob/master/src/interaction/AngularHarmonic.hpp#L102)

src/interaction/Cosine.hpp
```diff
@@ -1,107 +1,124 @@
...
+      real _computeForceRaw(real theta) const {
+
+        /*
+        real cos_theta = cos(theta);
+        if(cos_theta < -1.0) cos_theta = -1.0;
+        else if(cos_theta >  1.0) cos_theta =  1.0;
+
+        real sin_theta = sqrt(1.0 - cos_theta * cos_theta);
+        return K * sin_theta;
+        */
```
junghans: Just remove that block!

src/interaction/Cosine.hpp
```diff
@@ -1,107 +1,119 @@
...
+        }
+
+        
+        return true;
+      }
+      
+      // used for generating tabular angular potential
+      real _computeForceRaw(real theta) const {
+
+        return K;
```
MrTheodor: Is it correct to return constant for any theta?

acfogarty: No idea! Looks like Stas Bevc implemented it like that in commit 1a632760843e7b6f0589c4c22a14259cdcc67f59 (at the same time he was implementing everything else related to tabulated potentials, so I guess he knew what he was doing).

My first instinct was to put "_computeForceRaw(theta) not implemented in Cosine.hpp" in that function, but I'm not going to change a function I don't use or understand just because I'm doing a bugfix in another part of the file.

If you or someone else has some tested code to suggest please suggest it.

MrTheodor: It's really a hodgepodge so sure, this is not a case of this pull request.

I've briefly look on `TabulatedAngular.hpp` that should be a consumer of the ```_computeForceRaw```.
The value from table `A` is divided by `sin(theta)` and then used to compute force.

So I suppose that the `_computeForceRaw(theta)` should return `dU/dtheta = -K sin(theta - theta_0)` but... in `AngularHarmonic.hpp`, `_computeForceRaw` returns `-2.0*K*(theta - theta_0)/sin(theta)`.

PR #93, opened 2016-09-16
support for thermodynamic integration

acfogarty:
Code for doing free energies calculations using thermodynamic integration.
Includes non-bonded potentials (softcore Lennard Jones and linear Reaction Field) that are a function of a parameter lambda, and changes to the gromacs parser to use these potentials. Also a tutorial, example python script, and unit test for the new code.


src/tools/convert/gromacs.py
```diff
@@ -325,7 +325,14 @@ def read(gro_file, top_file="", doRegularExcl=True):
 
             if 'dihedraltypes' in line:
                 #is it really the dihedral (function type = 9) or is it actually the impropers (also labelled 'dihedraltypes' in gromacs but with function type = 4)
-                nextline = f.lines[lineindex + 2] #assumes one comment line between [ dihedraltypes ] and first parameters entry
+
+                #loop over any number of comment lines
+                ii = 1
+                nextline = f.lines[lineindex + ii]
+                while nextline[:1] == ';':
```
MrTheodor: or `f.lines[lineindex + ii].strip().startswith(';')` will do the same

src/tools/convert/gromacs.py
```diff
@@ -1065,6 +1072,55 @@ def setLennardJonesInteractions(system, defaults, atomtypeparams, verletlist, cu
     system.addInteraction(interaction)
     return interaction
 
+def setLennardJonesInteractionsTI(system, defaults, atomtypeparams, verletlist, cutoff, epsilonB, sigmaSC, alphaSC, powerSC, lambdaTI, pidlist, annihilate=True, hadress=False, adress=False, ftpl=None):
+    """ Set lennard jones interactions which were read from gromacs based on the atomypes"""
+    if (hadress and adress):
+      print "Error! In gromacs.setLennardJonesInteractionsTI, you cannot use adress and hadress at the same time"
+      return
+    if (hadress):
+        #interaction=espressopp.interaction.VerletListHadressLennardJonesSoftcoreTI(verletlist, ftpl)
```
MrTheodor: is it necessary to keep this commented code?

src/tools/convert/topology_helper.py
```diff
@@ -58,7 +58,12 @@ def FillFileBuffer(fname, filebuffer):
     for line in f:
 	if "include" in line and not line[0]==';':
 	    name=(line.split()[1]).strip('\"')
-	    FillFileBuffer(name, filebuffer)
+            try:
```
MrTheodor: wrong indentation?

src/tools/convert/topology_helper.py
```diff
@@ -58,7 +58,12 @@ def FillFileBuffer(fname, filebuffer):
     for line in f:
 	if "include" in line and not line[0]==';':
 	    name=(line.split()[1]).strip('\"')
-	    FillFileBuffer(name, filebuffer)
+            try: 
+ 	        FillFileBuffer(name, filebuffer)
+            except IOError:
+                #need to use relative path
+                name = fname.split('/')[0] + '/' + name
```
MrTheodor: What would you like to do here? If only to get the name of directory from fname then `os.path.join(os.path.dirname(fname), name)` should do the job

testsuite/thermo_integration/test_thermo_integration.py
```diff
@@ -0,0 +1,117 @@
...
+import sys
+import time
+import espressopp
+import mpi4py.MPI as MPI
+
+import unittest
+
+class TestThermoIntegration(unittest.TestCase):
+    def setUp(self):
+
```
MrTheodor: empty line

doc/ug/tutorial.ThermodynamicIntegration.rst
```diff
@@ -0,0 +1,157 @@
+Thermodynamic integration
+=========================
+
+.. |espp| replace:: ESPResSo++
+
+Theoretical explanation
+-----------------------
+
+Thermodynamic integration (TI) is a method used to calculate the free energy difference between two states A and B. For the theoretical background, see e.g. http://www.alchemistry.org. In this tutorial, we show how to perform TI calculations with ESPResSo++. We calculate the free energy of solvation of methanol in water. The complete python script is available in the ESPResSo+ source code under ``$ESPRESSOHOME/examples/thd_integration_solvation``
```
acfogarty: It's supposed to be the espressopp root directory. Actually if I just write "available in the ESPResSo+ source code under examples/thd_integration_solvation" I guess that's clear.

testsuite/thermo_integration/test_thermo_integration.py
acfogarty: What about it?

fedepad: I guess @MrTheodor  refers to the empty line between the first imports (sys, time, etc.) and "import unittest" otherwise I don't see a problem with it.

acfogarty: @fedepad I was thinking the same thing as you, but actually this file test_thermo_integration.py is copied directly from @MrTheodor 's test_minimize_energy.py, which also has all the same empty lines. I think he might be suggesting to add more empty lines before and after the class definition?!

src/interaction/LennardJonesSoftcoreTI.cpp
```diff
@@ -0,0 +1,86 @@
...
+      using namespace espressopp::python;
+
+      void (LennardJonesSoftcoreTI::*pyAddPid)(longint pid) = &LennardJonesSoftcoreTI::addPid;
+
+      class_< LennardJonesSoftcoreTI, bases< Potential > >
+    	("interaction_LennardJonesSoftcoreTI", init< real, real, real, real, real, real, real, real, bool>())
+        .def("addPid",pyAddPid)
+      ;
+
+      //class_< VerletListLennardJonesSoftcoreTI, bases< Interaction > >
```
acfogarty: I know we have a policy of not keeping chunks of old commented-out code, but I would argue this is different. It's for use in the future. If someone wants to use these potentials with VerletList or VerletListHadress (the two where it's most likely to be useful), then he just has to uncomment this code (and implement computeEnergyDeriv in the interaction template). It seems silly to make him figure out what to put and have to type it, when the code lines are already there.

src/interaction/LennardJonesSoftcoreTI.hpp
```diff
@@ -0,0 +1,281 @@
+/*
+  Copyright (C) 2012,2013,2016
```
acfogarty: ok

src/tools/convert/gromacs.py
acfogarty: Same comment as above on src/interaction/LennardJonesSoftcoreTI.cpp

testsuite/thermo_integration/test_thermo_integration.py
fedepad: @acfogarty Ah ok, I see it! Yes, that was also my other thoughts, maybe the empty lines before and after the class definition. Otherwise if it's the imports one could even imagine to agree on a convention to group the "system" imports (sys, time, mpi4py,unittest), then space, then import all the espressopp (user developed), as it's done sometimes in C/C++ code when someone wants to point out the system includes and the user developed includes or non-system/third parties includes, otherwise I don't see a problem with the way it is right now. But he will let you know what he meant!

src/interaction/LennardJonesSoftcoreTI.hpp
```diff
@@ -0,0 +1,281 @@
...
+	the softcore Lennard Jones potential
+        in a Thermodynamic Integration simulation
+    **/
+    class LennardJonesSoftcoreTI : public PotentialTemplate< LennardJonesSoftcoreTI > {
+    private:
+      real epsilonA, epsilonB;
+      real ff1A, ff2A;
+      real ff1B, ff2B;
+      bool annihilate; //if true, atoms in pidsTI are annihilated, otherwise they are decoupled
+      std::vector<real> pidsTI; //PIDs of particles whose LJ interaction is zero in TI state B
```
acfogarty: That's a mistake, should be `longint`. 

The length is typically N<100 particles but could in theory be much longer.

I just did some quick benchmarking for the worst-case scenario where the element is never found in the set/vector (which is what happens most of the time in typical applications of this class), looks like already when N>15 it's worth using set.

src/interaction/LennardJonesSoftcoreTI.cpp
govarguz: @acfogarty, adding a Tag on the commented code "for furhter functionality" or tagging it with "dev further functionality" will be useful. I found many former comments in the code with "TODO..." which was not self-explanatory

src/tools/convert/gromacs.py
govarguz: @acfogarty, adding a tag on the commented code "For Furhter Functionality" or tagging it with "dev further functionality" will be also useful. I found many former comments in the code with "TODO..." which was not self-explanatory

src/tools/convert/topology_helper.py
acfogarty: No, but the git diff displays wrong because there's a weird mix of spaces and tabs in that py file. Will clean it up a bit now (at least the part I touched!)

src/interaction/LennardJonesSoftcoreTI.cpp
acfogarty: @govarguz  ok, will do

src/tools/convert/topology_helper.py
acfogarty: Gromacs topology files give the path of other included topology files as filename or subdirectory/filename, relative to the cwd, but sometimes the file given simply as filename is actually in the subdirectory of the file it was included from, not in the cwd. Gromacs can handle this, espresso++ couldn't.

In Gromacs-format files I've only ever seen a relative path one subdirectory deep, hence the `fname.split('/')[0]`. I don't actually know how gromacs handles deeper directory trees. But it's probable it handles them the same way as your suggestion `os.path.join(os.path.dirname(fname), name)` so let's switch to that.

Plus it will work assuming we ever try to support Windows ;)

src/interaction/LennardJonesSoftcoreTI.hpp
acfogarty: Actually large portions of these files are based on the previously existing LennardJones and ReactionField. I'm not sure we should delete references to Frauenhofer.

testsuite/thermo_integration/test_thermo_integration.py
MrTheodor: @acfogarty Catch me ;) 
@fedepad yes, I meant about the blank line after the method name

src/tools/convert/topology_helper.py
MrTheodor: I did it a bit different:
```python
def FillFileBuffer(fname, filebuffer, cwd=None, defines=None):
    if cwd is None:
        cwd = '.'
    if defines is None:
        defines = {}
    f = open(os.path.join(cwd, fname), 'r')
    for line in f:
        if line.startswith(';'):
            continue
        if "include" in line:
            name = line.split()[1].strip('\"')
            cwd_name = os.path.dirname(name)
            if cwd_name != '':
                cwd = cwd_name
            FillFileBuffer(name, filebuffer, cwd, defines)
        elif 'define' in line:
            t = line.strip().split()
            if len(t) > 2:
                defines[t[1]] = ' '.join(t[2:])
        else:
            l = line.rstrip('\n')
            if l:
                filebuffer.appendline(l)

    f.close()
    return
```
it should work for opls and gromos

acfogarty: It doesn't seem to be working for me. If I run
`gromacs.read('conf.gro','/full/path/to/topol.top')`
or
`gromacs.read('conf.gro','topol.top')`
and topol.top contains the line
`#include "amber94.ff/forcefield.itp"`
and the file ./amber94.ff/forcefield.itp does exist, then `FillFileBuffer` (the version you suggest) goes looking for
`amber94.ff/amber94.ff/forcefield.itp`
What kind of directory structure are you using your version with? Also, I guess you also changed the line in gromacs.read() where FillFileBuffer is called. What did you put?

MrTheodor: I use it only with:
#include "/usr/share/gromacs/top/opls.ff/forcefield.itp"
in topol.top

this should help:
FillFileBuffer(os.path.basename(name), filebuffer, cwd, defines)

see here: https://github.com/MrTheodor/bakery/blob/master/src/gromacs_topology.py

acfogarty: Okay, got it.

The current version of topology_helper.py in this pull request does work with the line 
#include "/usr/share/gromacs/top/opls.ff/forcefield.itp"
in topol.top, including when forcefield.itp then includes other files whose path is given relative to `/usr/share/gromacs/top/opls.ff/`. And it also works for the various other cases I described above.

It doesn't work for `#define` statements, but that's always been the case. That would also require changes in gromacs.py. And it's not relevant to this pull request. (Actually I never use `#define` statements and don't know how they work :) )

acfogarty:
@MrTheodor Are you okay with the current version of the pull request, and if so can you remove your 'changes requested' review?

src/tools/convert/topology_helper.py
MrTheodor: #define are in GROMOS FF in GROMACS, that's why I need it

MrTheodor:
@acfogarty is it ready to merge?

acfogarty:
@MrTheodor From my point of view yes, it's ready to merge. If you want to do so that would be great.

PR #94, opened 2016-09-19
Fix printf format warning

fedepad:
Long int variables were formatted in printf as int. Just changed to the right formatting for that type and in this way removed the warning from Clang in OSX.

### DIFF
--- a/src/analysis/LBOutputScreen.cpp
+++ b/src/analysis/LBOutputScreen.cpp
@@ -1,6 +1,8 @@
 /*
-  Copyright (C) 2012,2013,2016
+  Copyright (C) 2016
       Max Planck Institute for Polymer Research & JGU Mainz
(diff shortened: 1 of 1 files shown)


fedepad:
@junghans I added only a letter to the format identifier in a printf. What's the problem with codecov/patch failing?

junghans:
The merge is effectually de-creasing the code coverage % and hence does the check fail.
I will merge it after you did a copyright update.

fedepad:
Should I just put 2016 in addition to the ones already there or should I add also add 2014,2015 ?

junghans:
Just add 2016

fedepad:
Just wait until the last commit ends in travis, since I had to fix the first copyright, the right date for JGU Mainz to be in (e.g. wasn't there in 2012).

PR #98, opened 2016-09-25
Reopen #32 - DumpXYZ modifications

fedepad:
Reopen #32 that was closed due to developer Github account refactoring.

Small and minimal extensions to DumpXYZ. Now can write standard and non standard XYZ files (e.g. as fastwritexyz). Printing of particle type is not hard coded anymore but retrieved and one can give the option to store pids and/or velocities (default set to standard XYZ format).
Precision for printing doubles set as fastwritexyx but can be changed. Changed fastwritexyz (had type hard coded too!) and others in DumpConfigurations.py accordingly.
One can now write:

type x y z (standard XYZ, minimal)
type x y z vx vy vz (standard XYZ, non minimal)
pid type x y z (non standard XYZ)
pid type x y z vx vy vz (non standard XYZ)

while before type was hard coded and one could write just
type x y z vx vy vz
without modification to the class plus precision was not set.

Proper tests will be provided, small local tests already performed.
Original DumpXYZ() was faster than fastwritexyz so this might be still faster too (to check) and if so maybe could be used instead of fastwritexyz() in src/io/DumpConfigurations.py.


niktre:
I know it may sound strange at 3 am, but here is my question: Did we really need a standard xyz? 

Wouldn't it better to provide, say, vtf standard as in old espresso-md?


Von meinem iPhone gesendet

junghans:
Just to clarify the spec (https://en.wikipedia.org/wiki/XYZ_file_format) says:
```
<number of atoms>
comment line
atom_symbol11 x-coord11 y-coord11 z-coord11
atom_symbol12 x-coord12 y-coord11 z-coord12
...
```
Everything else is any Espresso++-flavored version of xyz and you might risk that other tools like VMD or VOTCA cannot read it.

If you really want to create yet another file format go ahead, but maybe it would be better to use h5md, vtk, psf or even espressomd's block file format instead. The latter is like the xyz-format you are proposing, except it has a header line, which describes the columns, .e.g.
```
pos id x y z vx vy vz
title
data1 data2 ...
```

fedepad:
@niktre Few points here:
1) This was what was already there in the repo long before me, I didn't write the original class. I just extended a bit with the least changes to be able to fix the hard coded type of the particle and print pids if someone wanted that info. 
2) In this way, if it turns out to be faster than fastwritexyz(), it could be used instead of that.
3) The matter of supporting another file format such as VTF as you suggest it's not really a concern of these modifications but we can put either in the wiki or as a question/issue or we can just talk...
4) We can talk about VTF standard if this is more useful for someone!

fedepad:
@junghans I'm not proposing any new format...and if I would do it, it wouldn't be most probably in text format to store numbers...
1) This class was hard coding the particle type (which is required by the standard), which for me doesn't sound that right;
2) Default behavior is standard XYZ file format output (type x y z);
3) Users can ask (option) to print pids and velocities too. And yes, I agree this is not the standard XYZ format. But maybe some users need extra info and can write their own reader? Then why someone implemented writexyz or fastwritexyz for example? They know is not the standard (as documented by fastwritexyz_standard()) that some tools can support, but that doesn't mean they do not needed at that time, right?
4) I can revert all the changes and just keep the ones that concern retrieving and printing of particle type.

Need input.

src/analysis/ConfigurationExt.cpp
```diff
@@ -91,18 +91,51 @@ namespace espressopp {
...
 
-      int id = (*it).first;
+      RealND props = (*it).second;
+      return props;
+    }
+
+    void ConfigurationExtIterator::incrementIterator()
+    {
+      /*
+      if (it == end) {
```
niktre: why this block is commented out?

niktre:
@fedepad could you please change settings of you IDE regarding handling empty lines and spaces at the end of the lines? Otherwise it is very difficult to review the code...

junghans:
@niktre https://github.com/espressopp/espressopp/pull/98/files?w=1

niktre:
@junghans thanks!

src/analysis/ConfigurationExt.cpp
fedepad: This check is done also when the particle loop in DumpXYZ.cpp is entered, so i commented here, but obviously makes more sense to check at this point before the checks inside the loop for the next particle. My suggestion is to return a bool from ConfigurationExtIterator::incrementIterator() and have a check in DumpXYZ.cpp for that and exit the particle loop so there's no need for an extra comparison later. Thoughts?

src/analysis/ConfigurationExt.cpp
```diff
@@ -3,21 +3,21 @@
       Max Planck Institute for Polymer Research
```
junghans: Copyright update missing!

src/analysis/ConfigurationExt.hpp
```diff
@@ -3,21 +3,21 @@
       Max Planck Institute for Polymer Research
```
junghans: Copyright update missing

src/io/DumpXYZ.cpp
```diff
@@ -34,47 +38,151 @@ using namespace std;
...
+
+                } else {
+
+                  myfile << cei.currentProperties()[0] <<
+                  " " << fixed << setprecision(10) << cei.currentProperties()[1] << " " << cei.currentProperties()[2] <<
+                  " " << cei.currentProperties()[3] << " " << cei.currentProperties()[4] << " " << cei.currentProperties()[5] <<
+                  " " << cei.currentProperties()[6] << endl;
+                  myfile.unsetf(ios_base::fixed);
+                  myfile << setprecision(p);
+
```
junghans: Wouldn't it be cleaner to append to the current line  if `getStoreVelocities` or `getStoreVelocities` is set instead of duplicating everything?

src/io/DumpXYZ.cpp
```diff
@@ -34,47 +38,151 @@ using namespace std;
...
+          myfile << Li[0] * length_factor << "  0.0  0.0  0.0  "<<
                   Li[1] * length_factor << "  0.0  0.0  0.0  "<< Li[2] * length_factor;
           // additional info to comment line
           myfile << "  currentStep " << integrator->getStep() << "  lengthUnit "<< length_unit << endl;
-          
+
           ConfigurationExtIterator cei = conf_real-> getIterator();
-          if(length_factor == 1.0){
+          std::streamsize p = myfile.precision();
+          if(length_factor == 1.0){ // probably can remove this cond branch and
```
junghans: just set the length_factor to 1 by default and only implement one branch.

src/analysis/ConfigurationExt.cpp
junghans: Yeah, just drop it if the check has been moved.

src/io/DumpXYZ.cpp
fedepad: Will do!

src/io/DumpXYZ.cpp
fedepad: Done! The tests have passed, but please double check!

src/analysis/ConfigurationExt.hpp
fedepad: Done.

src/analysis/ConfigurationExt.cpp
fedepad: Done.

src/analysis/ConfigurationExt.cpp
fedepad: Dropped. Check last push.

fedepad:
Should I add the Copyright to testsuite/dump_xyz/test_dump_xyz.py ?

src/analysis/ConfigurationsExt.cpp
```diff
@@ -257,16 +275,26 @@ namespace espressopp {
...
+                                   ": " << coordinates[3*i] << " " <<  coordinates[3*i+1] << " " << coordinates[3*i+2] <<
+                                   "and velocities: " <<  velocities[3*i] << " " <<  velocities[3*i+1] << " " << velocities[3*i+2] <<
+                                   "and types: " <<  types[i]);
+
+
+             //RealND _vec(6);
+             RealND _vec(7);  // p[0].p[1].p[2].v[0].v[1].v[2].type
 
+             for (int k=0; k<1; k++)
+               _vec.setItem(k, types[i]);
```
MrTheodor: Does this for-loop make sense?
```C++
  _vec.setItem(0, types[i]);
```

src/analysis/ConfigurationsExt.cpp
```diff
@@ -257,16 +275,26 @@ namespace espressopp {
...
-             RealND _vec(6);
              for (int k=0; k<3; k++)
-               _vec.setItem(k, coordinates[3*i + k]);
+               _vec.setItem(k+1, coordinates[3*i + k]);
 
              for (int k=0; k<3; k++)
-               _vec.setItem(k+3, velocities[3*i + k]);
+               _vec.setItem(k+4, velocities[3*i + k]);
+
+             //for (int k=0; k<1; k++)
```
MrTheodor: Do you need this commented code?

src/analysis/ConfigurationsExt.hpp
```diff
@@ -1,23 +1,25 @@
 /*
+#  Copyright (C) 2016
+#      Max Planck Institute for Polymer Research & JGU Mainz
```
MrTheodor: copy-paste from Python code

src/io/DumpXYZ.cpp
```diff
@@ -1,27 +1,31 @@
 /*
+#  Copyright (C) 2016
+#      Max Planck Institute for Polymer Research & JGU Mainz
```
MrTheodor: another copy-paste

src/io/DumpXYZ.cpp
```diff
@@ -34,78 +38,93 @@ using namespace std;
...
-          else{
-            for(size_t i=0; i<num_of_particles; i++){
-            	//myfile << "bla" << endl;
-              myfile << "  0  " << length_factor * cei.nextProperties() << endl;
-              //myfile << "  0  " << length_factor * cei.nextCoordinates() << endl;
-            }
+          std::streamsize p = myfile.precision();
+          for (size_t i=0; i<num_of_particles; i++) {
+
+                if (getStorePids())  {
```
MrTheodor: Do you need to call those functions instead of directly using attributes?

src/io/DumpXYZ.cpp
```diff
@@ -34,78 +38,93 @@ using namespace std;
...
+                    myfile << cei.currentId() << " ";
+                }
+
+                myfile << cei.currentProperties()[0] <<
+                       " " << fixed << setprecision(10) << length_factor * cei.currentProperties()[1] << " "
+                       << length_factor * cei.currentProperties()[2] <<
+                       " " << length_factor * cei.currentProperties()[3];
+
+                if (getStoreVelocities()) {
```
MrTheodor: the same comment

src/io/DumpXYZ.hpp
```diff
@@ -80,28 +90,30 @@ namespace espressopp {
...
+
       std::string file_name;
-      
+
       bool unfolded;  // one can choose folded or unfolded coordinates, by default it is folded
       bool append; //append to existing trajectory file or create a new one
-      real length_factor;  // for example 
+      bool store_pids;
+      bool store_velocities;
+      real length_factor;  // for example
```
MrTheodor: "for example" what?

src/analysis/ConfigurationsExt.cpp
fedepad: No, makes no sense. I thought of that and then I kept the loop...will remove it!

src/io/DumpXYZ.cpp
fedepad: No need, I will switch to the attributes.

src/io/DumpXYZ.cpp
fedepad: I will switch to the attributes.

src/io/DumpXYZ.cpp
fedepad: True...

src/analysis/ConfigurationsExt.hpp
fedepad: True...

src/analysis/ConfigurationsExt.cpp
fedepad: Nope.

src/io/DumpXYZ.py
```diff
@@ -45,7 +47,15 @@
...
+    True if you want to store pids as fastwritexyz does. False otherwise (standard XYZ)
+    Default: False
+
+* `store_velocities`
+    True if you want to store velocities. False otherwise (XYZ doesn't require it)
+    Default: False
+
 usage:
 
 writing down trajectory
```
niktre: @fedepad line 66: change `xyz.dump()` to `dump_conf_xyz.dump() `

src/io/DumpXYZ.hpp
fedepad: Removed. Check last push.

src/io/DumpXYZ.py
fedepad: thanks @niktre !

fedepad:
@junghans Why https://travis-ci.org/espressopp/espressopp/jobs/164266478 fails?
I haven't touched the travis file...Looks like Sphinx and dependencies error...

fedepad:
I think all changes requested are there. Please double check!!!

src/analysis/ConfigurationsExt.cpp
```diff
@@ -257,16 +275,22 @@ namespace espressopp {
...
 
              LOG4ESPP_INFO(logger, "set coordianates of particle with id = " << index <<
-                                   ": " << coordinates[3*i] << " " <<  coordinates[3*i+1] << " " << coordinates[3*i+2] << 
-                                   "and velocities: " <<  velocities[3*i] << " " <<  velocities[3*i+1] << " " << velocities[3*i+2]);
+                                   ": " << coordinates[3*i] << " " <<  coordinates[3*i+1] << " " << coordinates[3*i+2] <<
+                                   "and velocities: " <<  velocities[3*i] << " " <<  velocities[3*i+1] << " " << velocities[3*i+2] <<
+                                   "and types: " <<  types[i]);
+
 
+             //RealND _vec(6);
```
junghans: Remove!

testsuite/dump_xyz/test_dump_xyz.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python
```
junghans: Yeah, I would add copyright here.

src/analysis/ConfigurationsExt.cpp
fedepad: Done. Check last push.

testsuite/dump_xyz/test_dump_xyz.py
fedepad: Done. Check last push.

Issue #51, opened 2016-02-07
mpi4py-2 breaks Espresso++


junghans:
We using a newer mpi4py version (0e356b00e444b4c550489bff54893c348e900e22) stuff breaks, see:
<https://travis-ci.org/espressopp/espressopp/jobs/107627162>

Error:
```
ValueError: too many values to unpack
```

fedepad:
@junghans I can confirm this failing test caused by that bug on OSX too with mpi4py version 2.0.0
I'll also have to mention that basically all tests fail on the same machine, so I don't know how they passed the Travis tests, even the recent ones, with the new mpi4py version (2.0.0)... 
We have to make sure no tests are failing with the old mpi4py (travis right?).
I refer for example to: FreeEnergyCompensation, HAdResS, ForceAdResS and thermo_integration just to mention the most recent ones.
I think I found the problem. 
Now all tests pass (except fixed_list_types_test, which was failing before too due to other reasons).
I will open a pull request, I want to discuss there some stuff about this.

junghans:
Travis installed mpi4py 1.3.1 using
```
pip install --user mpi4py==1.3.1
```

PR #100, opened 2016-09-30
Fix for issue #51: failing tests now pass with latest mpi4py (2.0.0)

fedepad:
Fix for issue #51 

With this modification all tests, except fixed_list_types_test, pass, while before, even the most recent ones (FreeEnergyCompensation, HAdResS, ForceAdResS and thermo_integration), where failing, at list with mpi4py version (2.0.0) on OSX.

I do not understand why the pmi.reduce() in Particle.py was returning two "values". Is it supposed to return two values or one? That was the problem basically, at least with the latest mpi4py version (2.0.0).

Was this working for earlier mpi4py versions? If so we could put an if in that part of Particle.py according to the mpi4py version.


junghans:
You will have to bump our bundled version of mpi4py in `contrib/mpi4py` as well.

fedepad:
@junghans thanks, I see it. Because it's running on my local machine but the tests fail when EXTERNAL=OFF.
BTW, should work with 1.3. now, since I check for the version, but I get that mpi4py doesn't have a __version__ attribute, which should be also present in that version. I don't get it.
But I will try to put the new version in contrib/ and will check that.

junghans:
Well, I would just drop support for 1.3 once 2.0 is included with `/contrib`.

fedepad:
do you have a clue what is happening for example in https://s3.amazonaws.com/archive.travis-ci.org/jobs/164144307/log.txt ?

This seems to be the error:

mpi4py_version = mpi4py.__version__.startswith('2')
5: AttributeError: 'module' object has no attribute '__version__'


But when I run on my machine (terminal):

>>> import mpi4py
>>> mpi4py_version = mpi4py.__version__.startswith('2')
>>> print mpi4py_version
True

So I don't know what's happening.

MrTheodor:
Are you sure that the mpi4py 2.0.0 is used? 1.3.1 indeed does not have
.version attribute;

fedepad:
@MrTheodor that's what it sounds like, since locally with mpi4py 2.0.0 I can retrieve the version the way I mentioned. I don't see what I'm missing since it seems to build the mpi4py 2.0.0 but it looks that when running the tests it doesn't use that. Pointers?

fedepad:
Update: in src/Particle.py there's no need to call pmi.MAXLOC
MAXLOC was one of the operation for which the reduce() op was changed. Since the rank is not returned and thus not needed, I replaced pmi.MAXLOC --> pmi.MAX and now tests seems to pass.
We should check if the call to pmi.reduce() in def node(self) in src/Particle.py should be modified; see
https://github.com/mpi4py/mpi4py/blob/master/CHANGES.rst

fedepad:
@junghans should we open a branch for this (updated to the latest master or update and use your new_mpi4py with this PR changes since your branch still has mpi4py 1.3.1 in contrib/, etc.) or maybe open a PR against the new_mpi4py branch and not against master or keep it as PR on master?
The choice is to the other developers and contributors.
Thoughts?

junghans:
Just merge 0e356b00e444b4c550489bff54893c348e900e22 into your branch and done.

junghans:
And please remove `contrib/mpi4py/mpi4py-1.3/`

fedepad:
Actually mine has already that version of .travis.yml to be able to install mpi4py 2.0.0 on travis.
And yes, I will remove contrib/mpi4py/mpi4py-1.3/

src/Particle.py
```diff
@@ -43,7 +43,7 @@
 from espressopp.Exceptions import ParticleDoesNotExistHere
 
 
-mpi4py_version = mpi4py.__version__.startswith('2')
+#mpi4py_version = mpi4py.__version__.startswith('2')
```
junghans: Drop this line

fedepad: Yes, I was going to do a last push without that line...

fedepad:
@junghans I just wait for any more input/changes request from you before doing another push.

junghans:
The rest:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 8bb2f53..5d2e8e3 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,10 +144,10 @@ include_directories(${CMAKE_SOURCE_DIR}/src)
 option(EXTERNAL_MPI4PY "Use external mpi4py" ON)

 if(EXTERNAL_MPI4PY)
-  find_package(MPI4PY 2.0 REQUIRED)
+  find_package(MPI4PY 1.3 REQUIRED)
   include_directories(${MPI4PY_INCLUDES})
 else(EXTERNAL_MPI4PY)
-  include_directories(${CMAKE_SOURCE_DIR}/contrib/mpi4py/mpi4py-2.0.0/src/include)
+  include_directories(${CMAKE_SOURCE_DIR}/contrib/mpi4py/mpi4py-1.3/src/include)
   set(MPI4PY_LIBRARIES MPI)
 endif(EXTERNAL_MPI4PY)

diff --git a/src/Particle.py b/src/Particle.py
@@ -1,5 +1,3 @@
-#  Copyright (C) 2016
-#      Max Planck Institute for Polymer Research & JGU Mainz
 #  Copyright (C) 2012,2013
 #      Max Planck Institute for Polymer Research
 #  Copyright (C) 2008,2009,2010,2011
@@ -38,13 +36,9 @@ import _espressopp
 import esutil
 import pmi
 from espressopp import toReal3DFromVector, toInt3DFromVector
-import mpi4py
-#import mpi4py.MPI as MPI
+import mpi4py.MPI as MPI
 from espressopp.Exceptions import ParticleDoesNotExistHere

-
-#mpi4py_version = mpi4py.__version__.startswith('2')
-
 # Controller Particle:
 # * requests are directly forwarded

@@ -198,5 +192,6 @@ if pmi.isController:
             return node

         def __getattr__(self, key):
-            value = pmi.reduce(pmi.MAX, self, 'getLocalData', key)
+            value, node = pmi.reduce(pmi.MAXLOC, self, 'getLocalData', key)
             return value
+
```
is trivial and look ok to me.

fedepad:
I will do the other push and see the builds.

fedepad:
BTW, very nice job with the command line magic!

fedepad:
Maybe we could change:
-  find_package(MPI4PY 2.0 REQUIRED)

to something >=1.3 otherwise may fail to find it if the system (external) mpi4py is <= 2.0,
while still this should work if mpi4py 1.3 for example.

Thoughts?

junghans:
mpi4py-2 got release Oct. 2015, I think we can consider that not too recent to make the default.
Beside the fact that we only support mpi4py 1.3.1 and 1.3 was broken, too.

@govarguz Any objections?

fedepad:
Should I add to the "Assignees" someone in particular?

PR #101, opened 2016-10-02
removed message that the DumpGRO destructor is called

MrTheodor:


acfogarty:
Great! That's been annoying me for ages. (There's just the copyright to update as usual)

PR #104, opened 2016-10-04
Reverted some of the changes in #98 inside ConfigurationsExt::gather() since they break current DumpGRO. Moved retrieval of particle type inside DumpXYZ

fedepad:
In DumpXYZ.cpp/.hpp reworked the retrieval of the particle type exactly as done in DumpGRO. 
The problem that I noticed now was that DumpGRO and some other classes rely
on config.nextProperties() and with my changes to ConfigurationsExt::gather() it would 
have printed a wrong line and required an ugly to see but very trivial fix in ConfigurationExtIterator::nextProperties().
Retrieving the particle type in this way instead, it still extend DumpXYZ with the addition of pids and velocity tracking, get the right particle type as DumpGRO and doesn't break a class now.

Should be ready to be included.


MrTheodor:
@fedepad why not to fix DumpGRO and other places where ConfigurationExt.cpp/ConfigurationsExt.cpp is used?

fedepad:
@MrTheodor That was my first idea. But there's one subtle thing that's easy to overlook and tests were still passing...that's why I would like to have more tests, like for DumpGRO too, etc.
In gather() I was building the RealND vector to push into the config. In this vector I was pushing the RealND vector <type,x,y,z,vx,vy,vz>. This way it's and would be super easy as one could just add the other remaining fields of the Particle struct eventually. I think you already see a problem:
type is an integer type, specifically should be an unsigned integer (size_t exactly), while RealND expects doubles. Tests were fine, etc. but I think that's not really correct even if the file produced look ok. You see what I'm saying?
BTW to have it work changing just a line in nextProperties()and keep the previous changes in gather() from #98  would take:
RealND coord_vel = RealND(6, &(*it).second); and return this. 
In this case DumpGRO behaves as expected in local tests I performed but I really do not recommend to go this way.

fedepad:
Added here the simple test about DumpGRO(), so next time there are some changes in some parts that hits methods used in that routine, we might be able to notice if it breaks or not. I will try to add later a test for the AdRes version of DumpGRO() so we can cover also that class (even if not in all corner cases).

fedepad:
Added simple test for DumpGROAdress() and organized the testsuite dir for classes in src/io in a way similar to the AdRes directory structure in the testsuite, I think it brings a little bit more organization to that part.

MrTheodor:
is it ready?

fedepad:
@MrTheodor I think so. Merged latest master and re-pushed. Tests were fine last time. Please re-check the changes and the tests when they finish. After that, should be ready.

fedepad:
Tests have finished, you can check!

PR #105, opened 2016-10-04
bugfix for inline RealND::RealND(const int _dim, const real v[3])

fedepad:
Array size set disregarding _dim and in the constructor could happen to access indexes out of the range for v[] (depending on the size of _dim).
If already _dim = 4 then there's a problem.

I think it's a bug but please confirm and have a look!!!


src/RealND.hpp
```diff
@@ -135,7 +137,7 @@ namespace espressopp {
       data[i] = v[i];
   }
   
-  inline RealND::RealND(const int _dim, const real v[3]) {
+  inline RealND::RealND(const int _dim, const real v[]) {
     setDimension(_dim);
```
MrTheodor: Nice one, Is this constructor use somewhere? 
Please add check up if number of elements in ```v``` is equal to ```_dim```.

fedepad: Temporarily I pushed a possible modification. Should write a test I think at some point for this in the spirit of the Array4DTest.

fedepad: I pushed simple tests for the constructors of RealND. Can put more test cases for the various methods; right now just for the constructors since the fix was in one of the constructor. We can talk about.

fedepad:
Tests run fine on my system which uses clang and the clang tests pass on Travis. I have to understand why the gcc's fail. Looks like a linking problem with MPI. I will look into that.

fedepad:
@MrTheodor I cannot find your comment in the commit or near by....but I will answer here.
I don't need it, but I don't understand why, clang builds and runs fine, while gcc does a mess with the link step for that executable.

The possible options are two and I tested both:
1) export in .travis.yml CC=mpicc CXX=mpiCC 
2) Change that CMakeLists.txt by adding the target_link_libraries(...)

MrTheodor:
@fedepad it seems to work, isn't it ? I've tested locally on gcc 5.4

fedepad:
@MrTheodor Yes, it looks like so. I've compiled on OSX, Linux with gcc 4.8.1 and on Travis and run the tests and it seems fine with gcc after those last changes.

fedepad:
@MrTheodor I can update with the most recent master, I was just waiting for #104 to be reviewed so that we get back DumpGRO() working again and keep DumpXYZ extensions before other stuff goes on top.

fedepad:
Merged latest master.

fedepad:
@junghans Thanks, I'll try that! Appreciated!

fedepad:
Merged latest master and re-pushed. Let's wait for tests.

fedepad:
Tests are over. Should be ready.

PR #103, opened 2016-10-02
new AdResS examples

acfogarty:
Examples scripts and input files for different protein+AdResS setups
These will also be referenced in @kkreis 's new AdResS tutorial

PR #106, opened 2016-10-04
new adress example scripts for peptides, proteins

acfogarty:
-protein with spherical atomistic region
-peptide with self-adjusting atomistic region composed of multiple
sphere

Addresses comments on closed pull request #103


junghans:
If the `itp` files are taken from Gromacs, we need to mention that in `/NOTICE`.

acfogarty:
@junghans /NOTICE ? Where do you mean exactly? e.g. as "copyright" headers in the itp files?

junghans:
Sorry, I mixed it up, Espresso is GPL-3 not ASL (which want a `NOTICE` files) so copyrights need to go into the files itself as Gromacs is LGPL-2.1 and hence different from GPL-3.

PR #109, opened 2016-10-04
AdResS tutorial added, some changes in (H-)AdResS examples

kkreis:
This adds the AdResS tutorial to the documentation and rearranges and modifies some of the (H-)AdResS examples. This is partly in preparation for the upcoming espresso++/votca workshop.


acfogarty:
How come several of the hadress examples are completely gone? Do the examples that remain really cover all the cases the old set of examples did?

kkreis:
Nothing is lost, it's the exact same stuff, just a bit better organised.

Before, there had been two more folders, I think. One of them was just the same as the "plain" system, but without the routines to measure pressure and density profiles. But the simulation was identical. Therefore, no need to have both. The other one did not have an actual simulation but only contained another tabulated potential that could be used as well as a python script to set up the tetrahedral molecules. But this script is available anyhow and from another tabulated potential we do not learn anything new... I think that the stuff is now much clearer and better organised.

acfogarty:
I'm going to accept this PR now so that @stuehn will include it in the version he's putting on USB sticks tomorrow for the espressopp school.  

If anyone has comments go ahead and make them anyway, and I guess @kkreis can make a second pull request at a later date. The comments just won't make it into the version used in the school...

kkreis:
@acfogarty Thanks for accepting.

As said already, I am happy to accept further comments, which I would then address in a later pull request.

PR #113, opened 2016-10-11
Updated polymer melt example description

govarguz:
Updated polymer melt example from the questioning in the Schloss Waldthausen tutorial


junghans:
Now #116.

PR #114, opened 2016-10-12
RATTLE for rigid bonds

acfogarty:
New integrator extension for constraining hydrogen-heavyatom bonds using RATTLE algorithm
Position and velocity constraints
Unit test


acfogarty:
In commit c3da5cc, I added a python tool which was referenced in the documentation in Rattle.py but which I forgot to include previously.

PR #117, opened 2016-10-13
ENH/PERF: Added the remove function FixedPairs and FixTriples Lists

govarguz:
Enhancement/Performance: Added the remove function FixedPairs and FixTriples Lists


MrTheodor:
@govarguz why not to do it in the destructor? also, name ```remove``` points out that method will remove something from the list and it would be nice to add the same interface to QuadrupleList.

govarguz:
@MrTheodor, I thought on the destructor, as well, but this is a "planned" PR which should handle a different functionality. The current remove function can be used from the scripting level down. The name is following the current E++ forms you can... remove "FixedLists" by its interactions "n".  QuadrupleList should also include such function. This will be included shortly...

src/FixedTripleList.py
```diff
@@ -88,6 +94,12 @@ def size(self):
         if pmi.workerIsActive():
             return self.cxxclass.size(self)
 
+    def remove(self):
+        if pmi.workerIsActive():
+            self.cxxclass.remove(self)
+            return
```
MrTheodor: why return None?

src/FixedPairList.cpp
```diff
@@ -257,6 +257,14 @@ namespace espressopp {
     LOG4ESPP_INFO(theLogger, "regenerated local fixed pair list from global list");
   }
 
+  void FixedPairList::remove(void) {
```
MrTheodor: you don't need to write (void), in C++ does not make any difference

src/FixedPairList.cpp
```diff
@@ -274,6 +282,7 @@ namespace espressopp {
       .def("add", pyAdd)
       .def("size", &FixedPairList::size)
       .def("getBonds",  &FixedPairList::getBonds)
+      .def("remove",  &FixedPairList::remove)
```
MrTheodor: I still think that the name ```remove``` does not represents what that method exactly do. Moreover, ```disconnect/connect``` and ```clear```/```clearAll```/```deleteAll``` are two different things. First is non-destructive whereas second removes information so in principle they should be separated.

MrTheodor: @govarguz 
If you run ```remove()``` defined here then this object cannot be used anymore because there is no way to connect to signals again.
For me, separating this to two different methods, one ```connect/disconnect``` and second ```deleteAll``` is a clean approach.
@niktre what do you think?

btw, I checked and calling ```del fpl``` on Python side does not call destructor...

govarguz: @MrTheodor, first of all, do not mix things in this PR. the precise idea is to remove (https://www.google.de/search?q=define:remove&ie=utf-8&oe=utf-8&gws_rd=cr&ei=BCoCWJWuLYr3aO3GuqAP) the FixedLists... so that a "newer" FixedList... can be created. The disconnect/connect+destructor will be my future PRs on this but this PR is only "remove"= cleanest approach. 
btw I believe there is no much discussion because of the "remove" term but I appreciate the feedback!

govarguz:
well the additional methods will be added, but additional PRs... this functionality is fulfilled and I am currently using it. However, as u mentioned a re-factoring with the additional PRs (functionalitites) can be added. @niktre any additional comment?

niktre:
@govarguz yes, my comment is: 
It is weekend. Relax till next working day! Breaks help thinking.

govarguz:
@niktre, thanks for reminding all that is not a working day! Indeed, sometime we might need to use a chat instead of PR comments for such things...

src/FixedPairList.hpp
```diff
@@ -67,6 +67,7 @@ namespace espressopp {
 
 	    python::list getBonds();
 	    GlobalPairs* getGlobalPairs() {return &globalPairs;};
+            void remove();
```
junghans: Indentation different from line above ....

src/FixedPairListAdress.hpp
```diff
@@ -58,8 +58,8 @@ namespace espressopp {
 		void beforeSendATParticles(std::vector<longint>& atpl, class OutBuffer& buf);
 		void onParticlesChanged();
 		python::list getBonds();
-
-		static void registerPython();
+                void remove();
+		static void registerPython();
```
junghans: Why are you adding whitespace errors, which @fedepad will have to fix later?

src/FixedTripleListAdress.hpp
```diff
@@ -58,7 +58,7 @@ namespace espressopp {
 		void beforeSendParticles(ParticleList& pl, class OutBuffer& buf);
 		void beforeSendATParticles(std::vector<longint>& atpl, class OutBuffer& buf);
 		void onParticlesChanged();
-
+                void remove();
```
junghans: Indentation again!

junghans:
And please check copyrights again!

govarguz:
ping @niktre

src/FixedPairListAdress.hpp
```diff
@@ -57,9 +57,9 @@ namespace espressopp {
 		void beforeSendParticles(ParticleList& pl, class OutBuffer& buf);
 		void beforeSendATParticles(std::vector<longint>& atpl, class OutBuffer& buf);
 		void onParticlesChanged();
+                void remove();
```
junghans: still spaces instead of tab here!

src/FixedTripleList.hpp
```diff
@@ -66,8 +66,9 @@ namespace espressopp {
 	    	return globalTriples.size();
 	    }
 
-		static void registerPython();
-	
+         void remove();
+                static void registerPython();
```
junghans: Space instead of tabs again!

src/FixedPairList.cpp
niktre: @MrTheodor I think that `remove` does exactly what it says, i.e. the functionality is correct. As far as I understand standard `connect` and `disconnect` are meaningless here (or at least are meaningless until somebody finds out why they may be used). @govarguz is this correct?

The question for me is whether we want to keep a uniform structure and preserve `connect` and `disconnect` in every module even though in some cases they make no physical sense. I do not have an answer here. Maybe @stuehn or @junghans can tell us what they think or how it supposed to be

govarguz: as I said: I am currently using remove, I personally don't see any application of connect disconnect YET! but if it is only to preserve structure could be added. I would say in the future connect disconnect could be added but with a functionality and thus within other PRs...

niktre: @govarguz suggesting you implement `connect/disconnect` and `removeAll` separately as suggested by @MrTheodor, will they have some identical functions inside (like `sigBeforeSend`, etc.)?

govarguz: @niktre, why would renaming <remove> to <removeAll> fix anything? It is not removing all fixedXLists, It seems to be misleading.... However I fixed the tabs @junghans asked for..

src/FixedPairListAdress.hpp
niktre: @govarguz make two tabs here

src/FixedPairListAdress.hpp
```diff
@@ -57,9 +57,9 @@ namespace espressopp {
...
 		python::list getBonds();
-
-		static void registerPython();
+                static void registerPython();
```
niktre: @govarguz make two tabs here

src/FixedTripleList.hpp
niktre: @govarguz make two tabs here for both lines

src/FixedTripleListAdress.hpp
niktre: @govarguz make two tabs here

PR #124, opened 2016-10-19
documentation fixes

acfogarty:
Added missing tools, interactions and integrator extensions to doc/ug

Made changes in python src files so that documentation displays correctly and looks consistent.


doc/ug/integrator.rst
```diff
@@ -25,7 +26,11 @@ integrator
    espressopp.integrator.LangevinBarostat.rst
    espressopp.integrator.LangevinThermostat.rst
    espressopp.integrator.LangevinThermostat1D.rst
+   espressopp.integrator.LangevinThermostatHybrid.rst
+   espressopp.integrator.LiquidGasLB.rst
```
niktre: Delete LiquidGasLB - it won't be there after #96 is accepted

src/interaction/LennardJonesSoftcoreTI.py
```diff
@@ -19,16 +19,17 @@
 #  along with this program.  If not, see <http://www.gnu.org/licenses/>. 
 
 
-"""
-***********************************************
+r"""
```
niktre: Was this file rendered wrong (missing/misspelled characters) without `r`?

doc/ug/integrator.rst
acfogarty: ok

acfogarty:
@niktre Mostly by hand, unfortunately. With grep and diff. That's why I only did integrator and interaction, since that's where I personally had added new classes with documentation but not added them to doc/ug. Someday soon (in a separate PR) I can write a simple script to do the same for the other subdirectories. 

The problem is, since many of the docstrings are not rendering correctly for various reasons, it's necessary to look at each individual html file anyway after building with sphynx, and modify them by hand. So you don't gain so much by just automatising creation of .rst files. In particular src/tools was a mess! (And still is, actually...)

Maybe there's somewhere in the wiki we can put a list of things people should do when they add a new class? As well as document, unittest, modify bindings and __init__, update copyright etc. it would also say add new .rst file in the appropriate place in doc/ug, build doc, and check the documentation they just wrote looks acceptable!

niktre:
@acfogarty you are right about wiki. I thought about it as well and will do this. But step by step :-)

niktre:
@acfogarty continuing the discussion on automatic generation of the `rst` files. There is `sphinx-apidoc` that in its default behavior generates missing `rst` files without overriding existing ones (at least it is claimed by `sphinx-apidoc --help`)

PR #132, opened 2016-11-15
Update docs structure and layout of the page

niktre:
This update changes the theme of the documentation to a modern one. It also brings the entries and the structure to a unified version (maybe not perfect, but it is definitely a start). Introduced grouping into several rst files


doc/ug/conf.py
```diff
@@ -46,10 +46,9 @@
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 
-extensions = ['sphinx.ext.pngmath',
-              'matplotlib.sphinxext.only_directives',
-              'matplotlib.sphinxext.plot_directive',
+extensions = ['sphinx.ext.mathjax',
               'sphinx.ext.autodoc',
+              'sphinx.ext.napoleon',
```
acfogarty: Is this really a good idea? I know the Google Python docstring style is much nicer than the standard sphinx rst one, but for the moment all our docstrings are written in the standard sphinx rst one anyway, and this is yet another dependency that the user may not have (doesn't seem to be on MPIP desktops, for instance)

doc/ug/conf.py
```diff
@@ -128,6 +127,7 @@
 # Sphinx are currently 'default' and 'sphinxdoc'.
 #html_theme = 'basic'
 #html_theme = 'alabaster'
+html_theme = 'sphinx_rtd_theme'
```
acfogarty: Same comment as above - it's another dependency that the user may not have. Is it really so much better than basic?

doc/ug/conf.py
niktre: As of Sphinx 1.3, the napoleon extension will come packaged with Sphinx under `sphinx.ext.napoleon`. The `sphinxcontrib.napoleon` extension will continue to work with Sphinx <= 1.2.

On my desktop (`pckr192`) I have `Sphinx (sphinx-build) 1.2b3`. What about you?

doc/ug/espressopp.rst
```diff
@@ -23,10 +22,13 @@ espressopp
    espressopp.Particle.rst
    espressopp.ParticleAccess.rst
    espressopp.ParticleGroup.rst
+   espressopp.pmi.rst
    espressopp.Real3D.rst
    espressopp.RealND.rst
    espressopp.Tensor.rst
+   espressopp.System.rst
```
niktre: Yes, I try to restore alphabetical order everywhere

niktre:
I can revert to the old formatting for now, sure. However, we probably have to find some way to structure the `tools` in a way that makes sense. And maybe also merge it with `io`...

niktre:
As for the second point, we can discuss it further tomorrow or next week. 

In short: you are right one has to be able to compile the docs at MPIP desktop and I will ask Robert to update Sphinx to 1.4.something. On the other hand I think you are wrong about 

> We won't get MPIP people helping with documentation if they...

It should read `We won't get MPIP people helping with documentation at all.` And in this case it is not a very high barrier to install a newer sphinx for two of us (and maybe Karsten).

doc/ug/conf.py
niktre: I think a user should not compile docs by definition

niktre:
@acfogarty I've moved the files from `tools/convert` to `tools`. However there is one problem:
there are some files (in `testsuite` and `examples`, for example `examples/polymer_melt/polymer_melt.py` or `testsuite/pi_water/water.py` or some `examples/fadress` files) where one does:

`from espressopp.tools.convert import gromacs`

Of course, this does not work anymore and the corresponding unittests fail. What do you think is better: 
1. Move files back to `convert`

2. Modify the `import` lines to account for a new path to "converting" tools?

niktre:
I'm almost finished. Need only to update the layout of docs...

niktre:
Ok, now I'm done!

acfogarty:
Looks good!

But have you tried compiling this on a few different MPIP desktops? Which ones? If it's not easy to compile on MPIP desktops, we can't insist people document... I get various problems. The main one is:

conf.py is still using sphinx_rtd_theme, which causes an error on MPIP desktops. Why not switch back to basic or default?

Also, various conflicts between Python 2 and 3 in doc/ug/sphinxext/sphinxtogithub.py and with pickle (do we really need the pickle option in the Makefile?) But I see you didn't touch these files, so I guess it's a problem of the desktops, not this PR.

Anyway, I'd like to know which desktops you CAN compile on before tracking down the problems further.

niktre:
I compiled on my desktop, pckr192, and on Saeed's (have no idea what name it is). It worked. What name your desktop has? Maybe I will try to compile it from there?

I haven't touch sphinxtogithub. But I think @junghans wanted to remove this plugin in some other PR...

junghans:
Done in espressopp/espressopp.github.io@c3a6c7314da10bea3c0d94c779dcfb46043bb5f3 and 948aff7590d9da6cb52f0a1dc33ad7916e734936.

junghans:
On a related note, http://www.espresso-pp.de/ should be forwarded to http://espressopp.github.io.
The latter is updated automatically from master (see #42, #43 and #55).

acfogarty:
> http://www.espresso-pp.de/ should be forwarded to http://espressopp.github.io.

The documentation at http://www.espresso-pp.de/Documentation/classdescription.html corresponds to the tar file at http://www.espresso-pp.de/Documentation/instalation.html (@stuehn can confirm this?)

But http://espressopp.github.io/instalation.html tells to user to download a tar from https://github.com/espressopp/espressopp/releases, and now there are two problems:
1. This page hasn't been updated since Nov 13, 2014
2. Even if we update the page fairly often with a new tar, there's a bigger problem: The documentation at http://espressopp.github.io/classdescription.html is updated automatically from master, i.e. it corresponds to the current developer version, not to any particular tar file.

That's why we have two documentation websites. One for the developer version, constantly changing. One for the tar file containing a stable release version. If we redirect http://www.espresso-pp.de/, where do we put the documentation for the stable release version?

I know we discussed switching to a system where we have a release branch and a development branch, instead of the tars and master as now, but even if we do that, we still somehow need a system where we're hosting two different versions of the documentation.

junghans:
1.) Well it seems somebody forgot to push the git tag (#138) for v1.9.4, github will create a tarball automatically. 

2.) You could deploy a version of http://espressopp.github.io/classdescription.html for each tag  using travis easily and put it in a `v1.X` subfolder and link it on the main page.

niktre:
1. I know who this "somebody" was, but will not tell it :-) It was not actually a "proper" update, but an update for tutorial organized in October.

2. I suggest to separate master (stable) and development (unstable) branches as we were thinking for a long time.

junghans:
1.) See #139, I found the version!
2.) Let's merge #134 and I will take a crack at it.

junghans:
For 2.) see #140

acfogarty:
@niktre Could you try compiling the new docs on pckr197 or pckr154? And check if your user profile is sourcing any python or sphinx-related stuff that the average user wouldn't?

niktre:
I can compile without problems on `pckr197`. In my `.bashrc` there are following lines:

`
export PYTHONPATH=$PYTHONPATH:/usr/lib64/python2.7
export LD_LIBRARY_PATH=/usr/lib64:$LD_LIBRARY_PATH
export LD_LIBRARY_PATH=/usr/lib64/python2.7/site-packages/mpi4py:$LD_LIBRARY_PATH
export LD_LIBRARY_PATH=/usr/X11R6/lib64:$LD_LIBRARY_PATH
`

niktre:
@junghans by the way, how did you commit to my pull request?

junghans:
Seems like you allowed maintainer commits (checkbox on the right) to your branch.

niktre:
@acfogarty have you checked your `.bashrc`? Is yours similar to mine?

acfogarty:
@niktre I suppose the real question is what the .bashrc of the standard user at MPIP looks like, and he  doesn't have those lines (and they're also not under /etc).  Anyway, I just added those lines to my bashrc, but it doesn't help with finding sphinx_rtd_theme. I think you are sourcing sphinx_rtd_theme some other way?

niktre:
no, I don't. This theme is found by `conf.py`

niktre:
I also do not understand yet why a standard user of MPIP should be able to build the ug...

acfogarty:
Yeah, sorry, 'standard user' is not the right expression. I'm thinking of people like @jsmrek, Patrick, Marc, Elio, Hsiao-Ping, Chan, Bhaskar, Manjesh, etc. who may make occasional contributions to espressopp. I think it was decided everyone should have to document all new contributions before they can be accepted, but if people get stuck at the very first step (compiling documentation) then I'm afraid they will get annoyed and not do anything...

If it works for everyone except me then I will withdraw all my comments ;)

niktre:
@acfogarty ok, I see your point. Though in my opinion we can ask for documenting new stuff in a manual mode: as soon as new code appear we explain how to make documentation.

@govarguz and @kkreis could you please try to compile documentation of this pull request on your machines? I do not understand why it doesn't work for @acfogarty

junghans:
The website now contains multiple version of the user documentation, e.g. http://espressopp.github.io/v1.9.4.1/classdescription.html
(V1.9.4.1 - Feel free to remove the tag and files again later!) vs. http://espressopp.github.io/classdescription.html (for master).

Adding links to side menu is something left for @niktre!

junghans:
Plus docker deployment:
https://hub.docker.com/r/espressopp/espressopp/tags/

PR #133, opened 2016-12-09
update README

niktre:
account for a safer (and faster) way to do cmake


junghans:
The fastest way to get Espresso++ is:
```
git pull espressopp/espressopp
git run -it espressopp/espressopp /bin/bash
```

README.md
```diff
@@ -15,12 +15,14 @@ QUICKINSTALL:
 
 ```
 # cd espressopp
-# cmake .
-# make
+# cmake -DEXTERNAL_BOOST=OFF -DEXTERNAL_MPI4PY=OFF .
```
junghans: Recommending this as the default is really a bad idea:
https://travis-ci.org/junghans/espressopp/jobs/182439306

On the lastest ubuntu our internal boost is too old for the shipped gcc.

README.md
```diff
@@ -15,12 +15,14 @@ QUICKINSTALL:
 
 ```
 # cd espressopp
-# cmake .
-# make
+# cmake -DEXTERNAL_BOOST=OFF -DEXTERNAL_MPI4PY=OFF .
+# make -j 4
```
junghans: And why 4? How about `make -j`?

README.md
junghans: Recommending this as the default is really a bad idea:
https://travis-ci.org/junghans/espressopp/jobs/182439306

On the lastest ubuntu our internal boost is too old for the shipped gcc.

README.md
niktre: I thought that modern computers have more than one CPU and `make -j 4` will be definitely faster than `make`. I've heard that sometimes one goes overboard if using simply `-j` but have no proof of it. Therefore I've added a comment later saying one can adjust the number. 

@govarguz  what do you think?

niktre:
what is `git run` ? never heard of it...

README.md
niktre: by the way, on espresso-pp.de we recommend:
```
If cmake . doesn't find all the libraries and include files use ccmake . to manually specify all search paths or to change compiler options.

If you have problems building ESPResSo++ with the system boost- and/or mpi4py- libraries, you can choose to use the ESPResSo++ contributed libraries with the command: (do a make clean before)
# cmake -DEXTERNAL_BOOST=OFF -DEXTERNAL_MPI4PY=OFF -DFFTW3_LIBRARIES=your_lib_path/libfftw3.a . 
# make
```

maybe we should use this form here as well?

README.md
junghans: How about `make -j -l$(nproc)` ?

junghans:
I meant docker, sorry!

README.md
niktre: that's great!

README.md
niktre: :-)))))))))))))

the good news is that they are installed.

the bad news is that they not always work, because of the heterogeneity of the machines, operating systems and other madnesses of the universe

govarguz:
@niktre it looks exactly as I suggested you! LGTM!

PR #140, opened 2016-12-14
.travis.yml: deploy tagged version as docker and website

junghans:


junghans:
@niktre : do want to make it part of #132 or merge it before?

PR #143, opened 2016-12-15
DOC/PERF: changes to make within README

govarguz:


README.md
```diff
@@ -26,7 +26,7 @@ QUICKINSTALL:
 ```
 # cd espressopp
 # cmake -DEXTERNAL_BOOST=OFF -DEXTERNAL_MPI4PY=OFF .
-# make -j -l$(nproc)
+# make -j -l$(nproc-2)
```
junghans: In bash:
```
$ make -j -l$(nproc-2)
bash: nproc-2: command not found
make: *** No targets specified and no makefile found.  Stop.
```

govarguz:
thnx! fixed the typo

README.md
```diff
@@ -26,7 +26,7 @@ QUICKINSTALL:
 ```
 # cd espressopp
 # cmake -DEXTERNAL_BOOST=OFF -DEXTERNAL_MPI4PY=OFF .
-# make -j -l$(nproc)
+# make -j -l$(nproc)-2
```
junghans: So how would that work for 1 and 2 core machines?

MrTheodor: What is $(nproc) ?

$ echo $nproc

Does not show anything on my machine...

junghans: I am still not convinced of the point of this change, using load=nproc-2 is as arbitrary as load=nproc.

Plus, when `make` parses the option using `atof()`, `atof(x-2)` becomes `atof(x)`, which effectively ignores the `-2`.

junghans: in bash:
```
make -j -l$(($(nproc)-2))
```
would work.

niktre:
One of the reasons to provide a smaller number than nproc consists in a VERY slow response of a machine upon compilation. I cannot even move the pointer on my mac and my desktop

Von meinem iPhone gesendet

govarguz:
@niktre exactly, that was the reason I would modify the bash line in a few moments!

niktre:
maybe it is better to write:

`# make [for multicore CPUs make -j #threads]`

?

junghans:
@niktre  Yeah, I think that is better!

Usually `make -j<cores+1>` is what I have seen used.

govarguz:
Combined all suggestions for changes to make @ README.md

PR #150, opened 2016-12-21
bugfix for DihedralHarmonic potential

acfogarty:
This PR fixes a few bugs in a dihedral potential. DihedralPotential assumed that all dihedral potentials in espressopp are symmetric around phi=0, but DihedralHarmonic is not. Also, the DihedralHarmonic class itself didn't take account of the fact that it's not symmetric around 0. It also didn't take account of periodicity.

PR also includes new unit test and documentation for DihedralHarmonic.

This PR breaks the unit test fixed_list_types_test, but I think that may be because the test is written implicitly assuming symmetry around phi=0. Can you check, @MrTheodor ?


acfogarty:
@junghans I know, as mentioned in my initial PR message above ;)

I need to figure out with the help of @MrTheodor (who wrote that failing unit test) what now needs changing in TabulatedDihedral. The unit test only gives potentials for positive phi values, maybe we should require that tabulated potentials cover the whole range from -180 to 180 degrees. Or alternatively, detect whether the user supplied -180 to 180 or 0 to 180, and react intelligently. This would also fix an issue I remember another user (Chan Liu) posting to the mailing list years ago, about asymmetric tabulated dihedral potentials.

I figured it would be easier to have this conversation if I went ahead and made the PR, even knowing the build fails.

junghans:
If only I could read ;-)

MrTheodor:
@acfogarty Hi, sorry for the delay... the test has to be changed with the same calculation of the angle (perhaps extracting calculation of phi on the level of DihedralPotential.hpp and use it as the function everywhere in dihedral potentials).
Maybe for the table, the  only assumption, which should be in the docs, is that the table has to be in the range -180, 180 deg?
GROMACS tables works in that range as well as LAMMPS input files.

src/interaction/DihedralHarmonic.hpp
```diff
@@ -187,13 +199,19 @@ namespace espressopp {
        *
        * @return The value of the force
        */
+
+      //TODO note: this function has not been tested and may need debugging
       real _computeForceRaw(real phi) const {
+        std::cout<<"Warning! The function _computeForceRaw(real phi) in DihedralHarmonic has not been tested and may need debugging"<<std::endl;
```
MrTheodor: Is it the same way as in the _computeForceRaw ?

acfogarty: Is _what_ the same way as in the _computeForceRaw? Phi calculation, you mean?

acfogarty:
@MrTheodor 

> Maybe for the table, the only assumption, which should be in the docs, is that the table has to be in the range -180, 180 deg? GROMACS tables works in that range as well as LAMMPS input files.

Agreed!

Ok, I will change the unittest, and see if there's a good way to separate out a calculatePhi function for use everywhere.

src/interaction/DihedralHarmonic.hpp
MrTheodor: Yes, the \phi convention; anyway is this _computeForceRaw used somewhere?

acfogarty: As far as I know, these _computeForceRaw(real) functions at the end of the most of the interaction potential files were intended for being called directly from python, e.g. if someone wanted to output a tabulated potential.

I never used any of them and don't know if anyone currently does... That's why I added the warning.

acfogarty:
In TabulatedDihedral I documented the fact that tabulated dihedral potentials should be supplied in the range -180 to +180. Because of the way values outside the range are handled in the Interpolation classes, I don't see a clean way to cause a crash if the user supplies a potential only in the range 0 to +180. Anyway, the forces will be the same whether the user supplies -180 to +180 or 0 to 180, only the energies are affected.

There was no unit test for TabulatedDihedral forces, so I made this check of forces using the current espressopp version (9018d8ef1683b1d60ef3299ca18de1817d142ab8) versus the version after this PR (6e3e5e6c18c574c4f4275b69d946ec517ca8745c):
https://www.dropbox.com/s/e6v13od0uaohqkb/pre150-versus-post150.png?dl=0
Forces are the same before and after #150 . The script is here: https://gist.github.com/anonymous/82930269c0f47f3088b26d015cea8947

I don't intend to refactor the code to have one single calculatePhi function.

Looking at TabulatedDihedral, I wonder whether it works with asymmetric potentials, but that's getting well outside the scope of this PR.

In summary, the code for this PR:
- fixes the bug in DihedralHarmonic
- all other analytical dihedral potentials work the same as before
- TabulatedDihedral works the same as before so long as the tabulated potential is supplied in range -180 to +180 (and this is documented)
- I think it's never been tested whether TabulatedDihedral works with asymmetric potentials (i.e. this PR doesn't change anything in that respect)

testsuite/interaction_potentials/dihedral_harmonic/test_dihedralharmonic.py
```diff
@@ -0,0 +1,130 @@
...
+    c[1]=u[2]*v[0]-u[0]*v[2]
+    c[2]=u[0]*v[1]-u[1]*v[0]
+    return c
+
+def dot(u,v):
+    c=u[0]*v[0]+u[1]*v[1]+u[2]*v[2]
+    return c
+
+def calc_dihedral(self,quadrupleslist):
+    #quadrupleslist contains [i,j,k,n]
```
MrTheodor: Where do you use this function?

testsuite/interaction_potentials/dihedral_harmonic/test_dihedralharmonic.py
```diff
@@ -0,0 +1,130 @@
...
+        fpl = espressopp.FixedPairList(self.system.storage)
+        fpl.addBonds(bondtuples)
+        interaction = espressopp.interaction.FixedPairListHarmonic(self.system,fpl,potential=espressopp.interaction.Harmonic(K=1.0,r0=1.0))
+        self.system.addInteraction(interaction)
+
+        integrator = espressopp.integrator.VelocityVerlet(self.system)
+
+        integrator.run(50)
+
+        self.assertAlmostEqual(interactions[0].computeEnergy(),0.747885,places=5)
```
MrTheodor: I suppose that the idea was to use it here as the check for computeEnergy calculated independently in the test?

testsuite/interaction_potentials/dihedral_harmonic/test_dihedralharmonic.py
acfogarty: Yes, as you say below, I was using the function calc_dihedral to output phi and independently check computeEnergy, and just generally check that the system was behaving as expected (by printing phi as a function of time over many steps). I forgot to delete it from the unit test before pushing. 

Actually now I'm reluctant to delete it because I spent some time checking that all definitions of phi were consistent (IUPAC convention, function calc_dihedral, convention used by espressopp in various places, convention used by vmd, convention used in gromacs, convention used in dlpoly...) and this can probably be useful for someone else in the future for debugging some other dihedral function. (You can see I also used it in the gist https://gist.github.com/anonymous/82930269c0f47f3088b26d015cea8947.)

I'll leave it in and add another assertAlmostEqual which uses it.

acfogarty:
As for using the gist as a unit test for TabulatedDihedral, it would certainly form a good basis for one, but I was reluctant to add a unit test for a class I never actually used myself.

acfogarty:
@junghans You left a 'request changes'. Are you okay with the PR now?

(Actually I have several PRs I must submit and get accepted this month before leaving MPIP, some of which depend tangentially on this one.)

Issue #149, opened 2016-12-20
Readme file


niktre:
Isn't it confusing that in #143 the sign `#` stays before the opening bracket (why do we need brackets there?) and note before `cores`?

junghans:
This is bikeshedding at its best! Let's just write `-j2` and be done with it!

niktre:
agree

acfogarty:
Any user who's confused by the `#` can just type `make` and be patient :)

On the other hand, I think some other parts of the README still have the potential to be confusing, e.g. now it looks like you have to install docker first. I propose #151

junghans:
I think `-j2` approximately halves the build time, while `-j3` only makes it drop from 50% to 33%, which only a bit better. Plus, espresso++ need a lot of memory to compile due to all the templates and hence `-j4` might be a bit risky on older machines. To me, `-j2` seems a reasonable compromise, and anyhow experienced users see the `-j` flag and will understand that we are supporting parallel builds and know what to do.

PR #151, opened 2016-12-21
more changes to README

acfogarty:
The most recent version of the README makes it look like the only way to get a working version of espressopp is via docker. Let's add info about git clone and the tarballs too. (I know those options are obvious to someone who's already using github, but it's not obvious to the potential user who googled his way to the espressopp github homepage.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ is free, open-source software published under the GNU General Public License
 QUICKSTART:
 ===========
 
-To get a copy of the developer version (most recent version) of ESPResSo++, you can use git or docker.
+To get a copy of the developer version (most recent version) of ESPResSo++, you can use git or docker. Using docker will give you a binary release (nothing to compile, but performance may not be optimal). If you use git clone or download a tarball, you will have to compile ESPResSo+ yourself, which might lead to better performance.
 
 Using [docker](https://www.docker.com):
 ```
@@ -34,7 +34,7 @@ QUICKINSTALL:
 ```
 # cd espressopp
 # cmake -DEXTERNAL_BOOST=OFF -DEXTERNAL_MPI4PY=OFF .
-# make [for multicore CPUs make -j#(cores+1)]
+# make -j2
 # source ESPRC
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,12 +13,20 @@ is free, open-source software published under the GNU General Public License
...
 
+Using git:
+```
+# git clone https://github.com/espressopp/espressopp.git
+```
+
+Alternatively, you can download a tarball or zip file of [previous release versions](https://github.com/espressopp/espressopp/releases) of ESPResSo++.
 
 QUICKINSTALL:
 =============
```
junghans: Could you include the fix for #149 as well?

README.md
```diff
@@ -32,6 +40,21 @@ QUICKINSTALL:
...
 After building go to the `examples` directory and have a look at the python scripts.
 
+DOCUMENTATION:
+==============
+
+Documentation for the developer version of ESPResSo++ is at:
+
+http://espressopp.github.io
+
+Documentation for release versions from v1.9.4.1 onward is at:
```
junghans: `v1.9.4.1` was more meant as a test release for the infrastructure hence I wanted to remove it later, but I am happy to keep it.

README.md
```diff
@@ -13,12 +13,20 @@ is free, open-source software published under the GNU General Public License
...
 
+To get a copy of the developer version (most recent version) of ESPResSo++, you can use git or docker.
+
 Using [docker](https://www.docker.com):
 ```
 # docker pull espressopp/espressopp
 # docker run -it espressopp/espressopp /bin/bash
 ```
 
+Using git:
```
junghans: Maybe we should mention that docker is binary release (noting to compile, but not optimal performance), while git is a compile it yourself approach, which might lead to better performance.

README.md
acfogarty: I put v1.9.4.1 as the example because it's the only alternative doc version at the moment, right? But in the future when presumably we will have several releases and corresponding docs we can put something else.

README.md
acfogarty: ok

README.md
acfogarty: ok

README.md
junghans: Good point, ok!

PR #153, opened 2017-01-26
Cosmetic changes to User Documentation

niktre:
Added images, logos. Updated icons. Removed some repeating text. Removed "Citations" menu item


acfogarty:
Looks great!

MrTheodor:
Wow awesome! btw, in the bottom of main page there is an img included "_static/MPIP_Logo.png" but it seems to not exists.

niktre:
@MrTheodor you are right. This picture however was successfully updated in 42ab329. It is probably happens due to the difference in capital letters. I'll change it right now, lets see if this works

junghans:
On a related note, doxygen and pdf manual links are missing. Plus, doxygen stuff is gone (fix niktre/espressopp#3).

niktre:
@junghans maybe I'm missing something but for me the [doxygen](http://www.espresso-pp.de/Documentation/doxygen/index.html) and [pdf manual](http://www.espresso-pp.de/Documentation/ESPResSo++.pdf) links are leading to nowhere on our website. That's why I removed both

junghans:
The pdf manual is here: http://espressopp.github.io/ESPResSo++.pdf
Doxygen got removed in espressopp/espressopp.github.io@610417a85a7bc31d0eac5fd1d284dddc1aa2f455, reason is the missing `;` which deletes it instead of deploying it (niktre/espressopp#3 fixes that).

PR #156, opened 2017-01-27
fix missing mpip logo

niktre:


junghans:
While you are at it, you might want to update the authors of the pdf manual.

niktre:
@junghans of course, thanks!

niktre:
@junghans now I'm stuck. Who do you think should be the authors? Everybody?

junghans:
Just put "Development team" or something like that.

niktre:
@acfogarty ok, now I added what I wanted. You only need to wait till all the checks are passed

PR #158, opened 2017-02-07
magical fixes and updates to the docs

niktre:
added sphinx_rtd_theme locally to the _themes folder (enables search field and removes dependence on installed theme);
removed old _templates;
visual adaptation of the layout (narrower side menu, clean look of the top left corner);
getting rid of pages that will be on the main website (faq, developer team and getting help)


junghans:
Where are the faq, developer team and getting help pages now?

niktre:
I made a redesign of the website, they are there. You can preview it at [www.espresso-pp.de/Secret/ ](url) but I'm still updating it, so it is not a final version yet.

acfogarty:
I don't completely understand. So you remove the list of developers, getting help and faq because they will be somewhere else on the site hosted at http://www.espresso-pp.de, when you switch to this new "Secret" design (it looks pretty cool, by the way!)

But what about the github.io site? Isn't that now going to lose these pages?

doc/ug/index_latex.rst
```diff
@@ -31,10 +31,6 @@ are different software packages.
    :maxdepth: 2
    :titlesonly:
 
-   overview.rst
    instalation.rst
    tutorial.rst
    classdescription.rst
-   FAQ <faq>
```
acfogarty: Aren't these now going to disappear from github.io too?

junghans:
Will this break the  old website trick?
```
http://espressopp.github.io/vXXX
```
where XXX is the version number, e.g.:
```
http://espressopp.github.io/v1.9.4.1
```

niktre:
Yes, the faq, getting help and developer team are gonna disappear from github.io. 
I thought we have to eliminate double pages. If a person finds documentation on github.io it can also find the developers (and faq, and getting help) on the main website. 
At the same time, if somebody comes to the main website and does not have any idea about the technical part, it is better to have faq, getting help and dev_team pages there.
We can change it if there is a better logic behind (and I do not deny this may be the case :-)), but I definitely want to exclude repetition of the same pages at two different places.

niktre:
@junghans A good question. I'm not sure. I do not understand the expression in the  `${TRAVIS_BRANCH} = @(master|v1.*)` in `.travis.yml` (line 53). How does it know about `v`?

junghans:
It is here https://github.com/espressopp/espressopp/blob/master/.travis.yml#L61, we simply deploy the html and pdf in a different folder.

niktre:
@junghans then I think it will be ok

junghans:
As it was hard in the past to update the main site, I would like to have the github.io page to stand by itself, i.e. having at least the team and faq there.

niktre:
@junghans don't forget there is also AUTHORS file in the repo (so, the 3rd developer list). 
We *can* add links to the dev_page, faq and getting help to the left-side menu or reincarnate overview page with links. However, there is a link to the main website at top left corner, so all the info is only 2 clicks away.
By the way, now I have a password to the server with the web, so it can be updated faster.

PR #159, opened 2017-02-08
two python helper functions

acfogarty:
two new python functions for:
- adding a standard set of restraints typically used in protein-ligand binding calculations
- calculating the reaction field energy associated with self and exclusion pairs, which is missing from the class ReactionFieldGeneralized (useful e.g. for comparing espressopp energies with energies from different MD packages)


src/tools/energy.py
```diff
@@ -0,0 +1,93 @@
+#  Copyright (C) 2016
```
junghans: 2017?

testsuite/python_tools/complex/test_python_tools_complex.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python
```
junghans: Copyright missing?

PR #163, opened 2017-03-04
refactoring of CMVelocity class

niktre:
includes:

- renaming of TotalVelocity to CMVelocity

- substituting ParticleAccess with AnalysisBase as a parent

- new unittest and updated documentation

### DIFF
--- a/doc/ug/analysis.rst
+++ b/doc/ug/analysis.rst
@@ -9,6 +9,7 @@ analysis
    espressopp.analysis.AnalysisBase.rst
    espressopp.analysis.Autocorrelation.rst
    espressopp.analysis.CenterOfMass.rst
+   espressopp.analysis.CMVelocity.rst
    espressopp.analysis.ConfigsParticleDecomp.rst
    espressopp.analysis.Configurations.rst
    espressopp.analysis.ConfigurationsExt.rst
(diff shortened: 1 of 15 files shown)


niktre:
@MrTheodor I'm pretty sure it was not supposed to work like this :-) what do you mean by separate units?

It is a first step in switching from `Observable` to `AnalysisBase` for all analysis tools (the later is essentially equivalent to the first one with one important exception: it is template-based, so valid for any datatype and therefore much more flexible). I just started with `TotalVelocity` class, as somebody was interested in it and the documentation was slightly confusing.

MrTheodor:
@niktre I thought about two routines, one that calculate the velocity and is in the ```analysis/``` and second somewhere in (```integrator/``` or ```storage/```) that modified velocity and can reuse calculation of the first one or internally creates and uses it. But maybe it is too much overhead for a such construction.

So currently, the correct approach for the new compounds in ```analysis``` is to use ```AnalysisBase``` as a base class ?

niktre:
@MrTheodor oh, I see. That indeed sounds better, but I guess it doesn't worth it at this stage, where so many other things exist.

Yes, new classes should be derived from `AnalysisBase`

MrTheodor:
For me looks good; Christoph?

PR #168, opened 2017-06-01
Fixed a bug and Deleted verbose message

hidekb:
I corrected a wrong potential form and deleted verbose message.

### DIFF
--- a/src/interaction/VSphereSelf.hpp
+++ b/src/interaction/VSphereSelf.hpp
@@ -121,7 +121,7 @@ namespace espressopp {
         real energy;
         energy  =   e1*pow(1.0L*M_4PI3*sigma2, 1.0L*mth)       \
                   + a1*NbNbNb/(sigma2*sigma2*sigma2) \
-                  + a22dNb*sigma2;
+                  + a2dNb*sigma2;
         return energy;
       }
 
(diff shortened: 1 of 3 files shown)


src/interaction/VSphereSelf.hpp
```diff
@@ -121,7 +121,7 @@ namespace espressopp {
         real energy;
         energy  =   e1*pow(1.0L*M_4PI3*sigma2, 1.0L*mth)       \
                   + a1*NbNbNb/(sigma2*sigma2*sigma2) \
-                  + a22dNb*sigma2;
+                  + a2dNb*sigma2;
```
MrTheodor: If I look on formula (http://espressopp.github.io/espressopp.interaction.VSphereSelf.html) this should be 2*a_2/Nb so this part was correct (or if not then please update VSphereSelf.py to make it consistent and preferable add reference to some source where this potential was firstly introduce)
Another issue, is this sigma squared ? again in the doc, there is only sigma_2 and not sigma^2

hidekb: According with "Flactuating soft-sphere approach to coars-graining of polymer melts, Soft matter, 2010, 6, 2282", sigma_2 in VSphereSelft.py means sigma^2.
Based on Eq. (4) in this paper, the coefficient of the term of sigma^2 is just not 2*a_2/Nb but a_2/Nb. Additionally, this potential form is consistent with the formula of forces described in VSphereSelf.hpp.
Hence, I will update VSphereSelf.py.

PR #174, opened 2017-07-07
XTC output class

GDeichmann:
I have been using this class in my local copy for a while now and thought it may be useful for the package.

Dumps the trajectory in the compressed binary .xtc format familiar from GROMACS.
Unit test is included, as well.


src/io/DumpXTC.cpp
```diff
@@ -0,0 +1,228 @@
+/*
+  Copyright (C) 2017
+      Gregor Deichmann (TU Darmstadt, deichmann(at)cpc.tu-darmstadt.de) 
+  Copyright (C) 2012-2016
```
junghans: This file didn't exist before 2017, so drop these lines.

src/io/DumpXTC.hpp
```diff
@@ -0,0 +1,173 @@
+/*
+  Copyright (C) 2017
+      Gregor Deichmann (TU Darmstadt, deichmann(at)cpc.tu-darmstadt.de) 
+  Copyright (C) 2012-2016
```
junghans: Same here.

src/io/libxdrf.c
```diff
@@ -0,0 +1,1008 @@
+/*
```
junghans: Should we really add this file???

Wouldn't it be better to use `find_package(libxdr)` to detect and external xdrfile library?

src/io/DumpXTC.py
```diff
@@ -0,0 +1,122 @@
...
+
+Both exapmles will give the same result: 200 configurations in trajectory .gro file.
+
+setting up length scale
+
+For example, the Lennard-Jones model for liquid argon with :math:`\sigma=0.34 [nm]`
+
+>>> dump_conf_xtc = espressopp.io.DumpXTC(system, integrator, filename='trj.xtc', unfolded=False, length_factor=0.34, length_unit='nm', append=True)
+
+will produce trj.gro with in nanometers
```
MrTheodor: shouldn't be `trj.xtc` instead of `trj.gro` ?

src/io/DumpXTC.py
```diff
@@ -0,0 +1,122 @@
...
+
+.. function:: espressopp.io.DumpXTC(system, integrator, filename, unfolded, length_factor, length_unit, append)
+
+		:param system:
+		:param integrator:
+		:param filename: (default: 'out.gro')
```
MrTheodor: or `out.xtc` ?

src/io/DumpXTC.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+        bool bOk = true;
+
+        int num=XTC_MAGIC,ival; //gromacs magic number (should be 1995)
+        float fval;
+        int i,j;
+
+        if( xdr_int(xdr,&num) == 0){
+            bOk = false;
+            return 1;
```
MrTheodor: please return `true`/`false` if the method return type is `bool`

src/io/DumpXTC.cpp
```diff
@@ -0,0 +1,228 @@
...
+      if( system->comm->rank()==0 ){
+        ConfigurationExtPtr conf_real = conf.back();
+        
+        int num_of_particles = conf_real->getSize();
+       
+        if(this->open("a")){ 
+        
+          ConfigurationExtIterator cei = conf_real-> getIterator();
+          //RealND::iterator ii;
+          //short ind;
```
MrTheodor: do you need this commented code?

src/io/DumpXTC.hpp
```diff
@@ -0,0 +1,173 @@
...
+        std::vector< std::map<long, short> > allParticleIDMaps;
+        boost::mpi::all_gather(
+            *getSystem()->comm,
+            myParticleIDToTypeMap, 
+            allParticleIDMaps);
+            
+        //merge all particle ID maps
+        for (std::vector< std::map<long, short> >::iterator it=allParticleIDMaps.begin(); it!=allParticleIDMaps.end(); ++it)
+        {
+          particleIDToType.insert(it->begin(), it->end());
```
MrTheodor: Do you use `particleIDToType` anywhere?

src/io/DumpXTC.hpp
```diff
@@ -0,0 +1,173 @@
...
+      std::string file_name;
+
+      //an array or an map where key: particle id and value: particle type
+      //we assume, that the type of a particle does not change over time
+      std::map<long, short> particleIDToType;
+      
+      bool unfolded;  // one can choose folded or unfolded coordinates, by default it is folded
+      bool append; //append to existing trajectory file or create a new one
+      real length_factor;  // for example 
+      std::string length_unit; // length unit: {could be LJ, nm, A} it is just for user info
```
MrTheodor: Do you save `length_unit` alongside with `xtc` file.

src/io/libxdrf.c
GDeichmann: I'd be happy to, but is there any such library that is accessible in a user-friendly manner? (as in: obtainable without installing the whole lot of gromacs, mdanalysis or vmd?)

MrTheodor: `write_xtc`, `read_xtc` functions are available in libgromacs.so so maybe it is possible to use them instead of attaching this file? They are declared here: `gromacs/fileio/xtcio.h`; not sure how difficult is to use it from outside.
With this, we will not have a problem of migrating any changes from GROMACS base code to our version of this library.
We could also add compilation flag to switch on this module which will set requirements on GROMACS..

CMakeLists.txt
```diff
@@ -77,6 +77,13 @@ find_package(FFTW3 REQUIRED)
 include_directories(${FFTW3_INCLUDES})
 
 ########################################################################
+#Process GROMACS settings
+########################################################################
+
+find_package(GROMACS REQUIRED)
```
junghans: I think Gromacs should be optional...

cmake/FindGROMACS.cmake
```diff
@@ -0,0 +1,22 @@
+# - Find GROMACS
```
junghans: Gromacs comes with cmake config files, so there is no need to write our own `FindGROMACS.cmake`, just use:
```
find_package(GROMACS CONFIG)
```

src/io/DumpXTC.hpp
```diff
@@ -0,0 +1,112 @@
...
+      bool getUnfolded(){return unfolded;}
+      void setUnfolded(bool v){unfolded = v;}
+      bool getAppend(){return append;}
+      void setAppend(bool v){append = v;}
+
+      static void registerPython();
+    
+    protected:
+
+      //static LOG4ESPP_DECL_LOGGER(logger);
```
MrTheodor: if you don't use it then please remove it

src/io/DumpXTC.cpp
```diff
@@ -0,0 +1,173 @@
...
+          ConfigurationExtIterator cei = conf_real-> getIterator();
+          Real3D *box = new Real3D [dim];
+          Real3D *coord = new Real3D [num_of_particles];
+          RealND props;
+          props.setDimension( cei.currentProperties().getDimension() );
+
+          //HACK: Only valid for orthorhombic BC. Will there be anything else in ESPP?
+          Real3D bl=system->bc->getBoxL();
+
+          for(int i=0;i<dim;i++){
```
MrTheodor: Why not to directly create `matrix gmx_box` here and pass it to `write()`, then you could remove for-loop in line 68.

src/io/DumpXTC.cpp
```diff
@@ -0,0 +1,173 @@
...
+            box[i][0] = 0.;
+            box[i][1] = 0.;
+            box[i][2] = 0.;
+
+            box[i][i] = bl[i]; 
+          }
+
+          for(int i=0;i<num_of_particles;i++){
```
MrTheodor: maybe here you could also directly create `rvec *gmx_x = new rvec[natoms];` and then you will not need for-loop in line 76; actually `write()` method seems to only rewrite data from one data structure to another one; perhaps this could be reduced to only `dump()` method ?

GDeichmann:
Changed the CMake files to look for the GROMACS config file.
Also added a define `HAS_GROMACS` that will trigger the compilation of gromacs functions. The `DumpXTC` class is essentially left empty if compiled without Gromacs library. 
I am not sure, though, whether this is the best solution. 

Another possibility would be to exclude the whole C++-Class from the build if a Gromacs library is not found. But then importing the io classes will trigger a call to the nonexisting `registerPython()` function. One would have to prevent the `__init__.py`  from importing the `DumpXTC` class (e.g. via an environment variable set in the ESPRC).

What are your thoughts on this?

CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ set(CMAKE_CXX_EXTENSIONS OFF) #...without compiler extensions like gnu++11
 # User input options                                                   #
 ########################################################################
 option(EXTERNAL_BOOST "Use external boost" ON)
+option(ENABLE_GROMACS "Use external gromacs library" OFF)
```
MrTheodor: minor, I would use name WITH_XTC

CMakeLists.txt
```diff
@@ -80,8 +81,12 @@ include_directories(${FFTW3_INCLUDES})
 #Process GROMACS settings
 ########################################################################
 
-find_package(GROMACS REQUIRED)
-include_directories(${GROMACS_INCLUDES})
+find_package(GROMACS CONFIG)
```
MrTheodor: This should be already in `if (ENABLE_GROMACS)` block, otherwise end-user will see beautiful CMake warning;

MrTheodor:
I think that the cleanest way is to switch off DumpXTC at the level of `io/bindings.cpp` instead of providing empty class with `#if statements` everywhere.
If you already have `HAS_GROMACS` flag then you can use it to switch off `DumpXTC::registerPython();` and `#include "DumpXTC.hpp"` in `io/bindings.cpp`

You will also need to remove `io/DumpXTC.cpp` from `ESPRESSO_SOURCES`. This you can do in `src/CMakeList.txt`, see e.g. lines 2 and 13;
something like
```
if (ENABLE_GROMACS)
  file(GLOB_RECURSE IO_XTC_SOURCE io/DumpXTC.cpp)
  list(REMOVE_ITEM ESPRESSO_SOURCES ${IO_XTC_SOURCE})
endif(ENABLE_GROMACS)
```
In `io/__init__.py` you can simply enclose `from espressopp.io.DumpXTC import *` with `try..except` block

```
try:
    from espressopp.io.DumpXTC import *
except ImportError:
    pass # or some message to the user
```

CMakeLists.txt
```diff
@@ -77,6 +78,20 @@ find_package(FFTW3 REQUIRED)
 include_directories(${FFTW3_INCLUDES})
 
 ########################################################################
+#Process GROMACS settings
+########################################################################
+
+if(WITH_XTC)
+find_package(GROMACS CONFIG)
+include_directories(${GROMACS_INCLUDE_DIRS})
```
junghans: This line needs to be move inside the `if(GROMACS_FOUND)`  block, otherwise you might add a `NOTFOUND` variable to the include directories.

CMakeLists.txt
```diff
@@ -77,6 +78,20 @@ find_package(FFTW3 REQUIRED)
...
+  if(GROMACS_FOUND)
+    add_definitions( -DHAS_GROMACS )
+  else(GROMACS_FOUND)
+    set(WITH_XTC OFF)
```
junghans: I am not sure if that can be done! `WITH_XTC` is an cached option, you would need to to use `set(...FORCE)` to do that. Plus, it is a bit confusing for the user, which explicitly enabled the option.

CMakeLists.txt
```diff
@@ -77,6 +78,20 @@ find_package(FFTW3 REQUIRED)
...
+  endif(GROMACS_FOUND)
+endif(WITH_XTC)
```
junghans: I guess, we could simplify it to:
```cmake
if(WITH_XTC)
￼  find_package(GROMACS CONFIG REQUIRED)
￼  include_directories(${GROMACS_INCLUDE_DIRS})
  add_definitions( -DHAS_GROMACS )
endif()
```
to enforce gromacs if (explicitly) enabled.

PR #177, opened 2017-07-24
Fixing the DPD thermostat

GDeichmann:
So far the implementtion of DPD in the `DPDThermostat` class is buggy.
Firstly, the standard (parallel) DPD part is wrong (the distance vector is used for projection, rather than the unit vector parallel to it).
Secondly, the transverse part is never called and the implementation does not look correct at all. 
Fixed the erroneous parts (to the best of my knowledge).


junghans:
@GDeichmann: Could you add a test please?

junghans:
Nice job!!!!

PR #182, opened 2017-08-04
Add new interaction class "ConstrainRG" for implementing hierarchical strategy

hidekb:
For implementing hierarchical equilibration strategy for polymer melts,
I added new interaction "ConstrainRG", which constrain the gyration radius of subchains as the desired value.
Additionally, the documentation and test python script for this class are added.


src/interaction/ConstrainRG.hpp
```diff
@@ -0,0 +1,93 @@
...
+	    void setK_rg(real _k_rg) {
+		k_rg = _k_rg;
+	    }
+	    real getK_rg() const { return k_rg; }
+/////////////////////////////////////////////////////////////////////////////////////////
+	    
+	    real _computeEnergy(real diff) {
+		
+		real energy = diff*diff;
+
```
MrTheodor: This seems to not correlate with force calculation. Can you put in the documentation, the eq. for the force and energy terms.

src/interaction/ConstrainRG.hpp
```diff
@@ -0,0 +1,93 @@
...
+		return k_rg*energy;
+	    }
+	    
+	    Real3D _computeForce(Real3D diff,
+				 real rg_id,
+				 real diff_rg,
+				 long unsigned int N_Constrain) {
```
MrTheodor: This looks like again the assumption that the mass is 1; I'm fine with it as soon as there will be a note in the documentation about this assumption.

src/interaction/FixedLocalTupleRgListInteractionTemplate.hpp
```diff
@@ -0,0 +1,448 @@
...
+		    if (rg.count(i)) {
+			RG[i] = rg[i];
+		    } else {
+			RG[i] = 0.;
+		    }
+		}
+
+		boost::mpi::all_reduce(*system.comm, RG, num_of_subchain, totRG, std::plus<real>() );
+		
+		if (system.comm->rank() == 0) {
```
MrTheodor: why only on rank 0?

src/interaction/ConstrainRG.hpp
hidekb: This potential form is based on "Equilibration of high molecular weight polymer melts: A hierarchical strategy, Macro Lett., 2014, 3, 198". I already write this information at the line 36 of ConstrainRG.hpp. Although, this potential form is correct, the formula of the force is wrong. Then I correct these.

src/interaction/FixedLocalTupleRgListInteractionTemplate.hpp
hidekb: This is a just mistake. Thank you for your suggestion. I removed this line.

src/interaction/ConstrainRG.hpp
hidekb: I miss that p->mass() multiply the force term. Hence, I correct this. Actually, the force term described in this function is wrong. Then I also correct this formula.
$U_{rg} = k_{rg} \left(R_{g}^2 - {R_{g}^{ideal}}^2\right)^2$,
where $R_{g}^2 = \frac{1}{N} \sum^{N}_{i} (\vec{r}_i - \vec{R}_c)$ and $\vec{R}_c = \frac{1}{N} \sum^{N}_{i}$. Note that $\vec{R}_c$ is not the center of mass but center of a sub chain.
Then the force is represented as
$F_{rg} = -4 k_{rg} \left(R_{g}^2 - {R_{g}^{ideal}}^2\right) \times (\vec{r}_i - \vec{R}_c)$.

PR #186, opened 2017-08-14
Fixed the bug for FixedLocalTupleCom and RG Interaction

### LINKED ISSUE #185: Problems when running ConstrainCOM in parallel
I had a try at the new ConstrainCOM feature. However, running in parallel seems to be problematic. With me, the code runs into a deadlock every time I initialize a new `FixedLocalTupleListConstrainCOM` object in python using multiple nodes.
I ran a short printf-debugging of the c++ constructor and it seems that `num_of_subchain` has different values on each node before being passed to `boost::mpi::all_reduce` in line 127 in FixedLocalTupleComListInteractionTemplate.hpp . 
@hidekb

hidekb:
Fix #185

When some CPU don't have any FixedLocalTupleList (FLTL), the fatal error arises.
Individual CPU has the different ``N_constrain``, namely ``N_constrain``
 becomes 0 or can't be determined at the CPU which don't have any FLTL.
Hence, the calculation is terminated abnormally or is into deadlock.
For removing this fatal error, I implemented this modification.


GDeichmann:
Tested the new version. Seems to resolve the issue I brought up (#185).

src/interaction/FixedLocalTupleComListInteractionTemplate.hpp
```diff
@@ -92,18 +117,20 @@ namespace espressopp {
...
-		for (FixedLocalTupleList::TupleList::Iterator it(*fixedtupleList); it.isValid(); ++it) {
-		    Particle *p = it->first;
-		    pid_list[pidK] = p->id();
-		    
-		    for(long unsigned int i = 0; i < pidK; i++) {
-			if (pid_list[i]/N_Constrain == pid_list[pidK]/N_Constrain) {
-			    std::stringstream msg;
-			    msg << "ERROR: Particle ID is redundant\n";
-			    err.setException(msg.str());
+		if (!fixedtupleList->empty()) {
```
MrTheodor: Why do you need everywhere to check if the list empty? isValid() on empty container should return `false` so the for-loop will not be executed at all.

hidekb: You are right. This if statement is not necessary. Then, I remove this and others similar like this.

PR #189, opened 2017-09-22
Generalized output from SystemMonitor and modified ExtAnalyze

jkrajniak:
- generalized SystemMonitorOutput
- added SystemMonitorOutputCSV : .csv output
- added SystemMonitorOutputDummy : print only values on the screen
- added support for the vector values from the Observable
- minor fix in System.py
- ExtAnalyze calculates properties exactly every n*interval simulation steps


MrTheodor:
Completely do not know why LBDOutput is broken. It is related to ExtAnalyze. Basically, the old approach uses `counter`, the properties were calculated at simulation steps: 1, interval+1, 2*interval+1, ...

junghans:
@stuehn do we still need this?

jkrajniak:
I'll try to adjust this PR to the recent changes.

src/analysis/Observable.hpp
```diff
@@ -69,7 +69,7 @@ class Observable : public SystemAccess
     /** returns observable of type int, used for Python and on C++ level*/
     virtual int compute_int() const { return 0; };
     /** computes vector of real values (e.g. pressure tensor, ...), used on C++ level only */
-    virtual void compute_real_vector() { return; };
+    virtual std::vector<real> compute_real_vector() { return std::vector<real>(); };
```
XzzX, line 72: Is throwing a not implemented error a better default?

testsuite/lattice_boltzmann/test_LBMDcoupling.py
```diff
@@ -177,11 +177,11 @@ def test_lbmdcoupling(self):
...
         print("total LB-MD mom:  ", ("{:>18.1e}"*3).format(*tot_mom), "\n")
-
-        # momentum checks
+    #
+    #     # momentum checks
         self.assertAlmostEqual(tot_mom[0], 0., places=10)
-        self.assertAlmostEqual(tot_mom[1], 0., places=10)
-        self.assertAlmostEqual(tot_mom[2], 0., places=10)
+    #     self.assertAlmostEqual(tot_mom[1], 0., places=10)
+    #     self.assertAlmostEqual(tot_mom[2], 0., places=10)
```
XzzX, line 184: Why are these commented out? Either remove or make them work?

testsuite/lattice_boltzmann/test_LBMDcoupling.py
```diff
@@ -209,9 +209,9 @@ def test_restartlbmd(self):
         print("total LB-MD mom:  ", ("{:>18.1e}"*3).format(*tot_mom), "\n")
 
         # momentum checks
-        self.assertAlmostEqual(tot_mom[0], 0., places=10)
-        self.assertAlmostEqual(tot_mom[1], 0., places=10)
-        self.assertAlmostEqual(tot_mom[2], 0., places=10)
+        # self.assertAlmostEqual(tot_mom[0], 0., places=10)
```
XzzX, line 212: Why are these commented out? Either remove or make them work?

src/analysis/SystemMonitor.hpp
```diff
@@ -35,15 +35,44 @@
...
+/**
+ * Base class for SystemMonitor output formats.
+ */
+class SystemMonitorOutput
 {
     friend class SystemMonitor;
 
+public:
+    virtual void write() = 0;
+    void setSystem(shared_ptr<System> system) { system_ = system; }
```
XzzX, line 53: Not really important, more matter of style. This will introduce an unnecessary copy, you could use const&.

XzzX (approved):

testsuite/lattice_boltzmann/test_LBMDcoupling.py
jkrajniak: Thanks, reverted

testsuite/lattice_boltzmann/test_LBMDcoupling.py
jkrajniak: Thanks, reverted

src/analysis/Observable.hpp
jkrajniak: I would keep it as it is - other virtual method of this class returns already default values. I'm not sure what will break if I throw an exception here.

src/analysis/SystemMonitor.hpp
jkrajniak: I think here we should copy the shared_ptr to share the ownership.

XzzX: Yes, of course, keep the shared_ptr. I meant you are copying the shared_ptr twice. Not relevant for a shared_ptr but relevant if you copy something like std::vector or std::string.

jkrajniak: So it's better to use:
```suggestion
    void setSystem(const shared_ptr<System> &system) { system_ = system; }
```
?

PR #193, opened 2017-09-25
Speed up compilation

MrTheodor:
- added ccache option to CMake
- gitversion can be switched off


MrTheodor:
The advantage is that it works out-of-box without additional effort.
I've just put `/usr/lib/ccache` in my PATH but this did not give any compilation speed-up. I suppose that I have to do something more than that on Ubuntu 17.04;

junghans:
We had a similar feature in VOTCA, but it led to a lot of pain with broken compilers unsupported by ccache and hence we dropped it again.

Plus, the `RULE_LAUNCH_*` isn't support by the Ninja generator.

MrTheodor:
Ok, if this cause problem with other compilers then I will look for some local solution. I simply got a bit annoyed with recompilation again and again when I checkout into a new branch.

PR #194, opened 2017-09-25
Update of TDforce routine

MrTheodor:
- switch from std::map to std::unordered_set (lookup is constant (average) instead
of logarithmic)
- addForce allows reading again the table with the new one when it is
called again with the same type


MrTheodor:
@hidekb could you review it

PR #197, opened 2017-09-27
Declutter the code base

### LINKED ISSUE #39: Clean up dirty code base

jkrajniak:
Fix #39 

- added new target (`make format`) which basically runs `clang-format -style=Google`
- run the command on all files

It's not perfect, but I think better than nothing. But all tests works
Next steps are to clean according to PEP8 .py files and add in travis check for the code style.


niktre:
this looks very good to me

junghans:
It would be nice to have travis run it, too and fail it `make format` change any files (test by e.g. `git status`.)

MrTheodor:
@junghans perhaps we could merge it with #201

Issue #201, opened 2017-10-08
Set travis to check if the code in pull request is formatted correctly


jkrajniak:
Travis should check the files in the pull request (only those files) and if any of the files do not follow the code style then the build should be marked as broken.
Preferably, this should be done before doing the compiling&testing phase (travis staging? )

Get the list of files from pull request 194:
```
curl -s "https://api.github.com/repos/espressopp/espressopp/pulls/194/files" | grep '"filename":' | while read a; do echo ">${a}"; done
```
or with `git diff` 
```
git diff --name-only $TRAVIS_COMMIT_RANGE
```

cc: #197

PR #202, opened 2017-10-10
Set travis to check code style before accepting pull request

jkrajniak:
- added staging in travis, first check code style then run the tests
- added `.sh` script to run `clang-format` on files that are in the pull request and check if anything requires changes (not auto changes)
- this should be extended to run flake8 or pylint on .py files

Remark, only code style is checked and not the code itself, this could be done e.g. by cpplint

closes: #201 
requires: espressopp/buildenv#4


maintainer/check_codestyle.sh
```diff
@@ -0,0 +1,39 @@
...
+        if [[ "${fchanged}" =~ .*\.[ch]+pp ]]; then
+            clang-format -style=file ${fchanged} | diff -q - ${fchanged} &> /dev/null || INCORRECT_FILES="${INCORRECT_FILES} ${fchanged}"
+        fi
+    done
+}
+
+#function pep8format() {
+#}
+
+if [ "X$CODESTYLE" = "XON" ]; then
```
junghans: if it is bash, which it is. Use `[[ $CODESTYLE = ON ]]` - no quotes needed.

docker/Dockerfile
```diff
@@ -11,12 +11,15 @@ ENV CI true
...
-ARG TRAVIS_PULL_REQUEST 
+ARG TRAVIS_PULL_REQUEST
 ARG TRAVIS_JOB_ID
 ARG TRAVIS_TAG
 ARG TRAVIS_REPO_SLUG
+ARG CODESTYLE
 ARG TRAVIS_COMMIT
+ARG TRAVIS_COMMIT_RANGE
 
+RUN [[ ${CODESTYLE} = "ON" ]] && exit 1
```
junghans: It there are chance to put everything one dockerfile?

Just reorder the lines below and `exit` before `ccache` and the build.

maintainer/check_codestyle.sh
```diff
@@ -0,0 +1,39 @@
...
+    if [ "X$TRAVIS_PULL_REQUEST" != "Xfalse" ]; then
+        # Gets list of files from pull request.
+        FILES="`curl -s "https://api.github.com/repos/${TRAVIS_REPO_SLUG}/pulls/${TRAVIS_PULL_REQUEST}/files" | grep '"filename":' | grep '"filename":' | sed -e 's/.*"filename": "\(.*\)",/\1/g' | xargs`"
```
junghans: Couldn't we do that with git instead, something along the line of:
```
git diff --stat ${TRAVIS_BRANCH}..HEAD
```

docker/Dockerfile
MrTheodor: I set a single dockerfile. If everything is fine it will continue to build and run the tests.

maintainer/check_codestyle.sh
MrTheodor: Changed

maintainer/check_codestyle.sh
MrTheodor: Ok

maintainer/check_codestyle.sh
```diff
@@ -0,0 +1,28 @@
...
+function clangformat() {
+    for fchanged in $1; do
+        if [[ "${fchanged}" =~ .*\.[ch]+pp ]]; then
+            clang-format -style=file ${fchanged} | diff -q - ${fchanged} &> /dev/null || INCORRECT_FILES="${INCORRECT_FILES} ${fchanged}"
+        fi
+    done
+}
+
+#function pep8format() {
+#}
```
junghans: remove...

maintainer/check_codestyle.sh
```diff
@@ -0,0 +1,28 @@
...
+
+if [ $CODESTYLE = ON ]; then
```
junghans: use `[[ .. ]]`

maintainer/check_codestyle.sh
```diff
@@ -0,0 +1,28 @@
...
+    FILES="`git diff --name-only --diff-filter=AM $TRAVIS_BRANCH...HEAD`"
```
junghans: use `$(..)` instead of backqoutes.

maintainer/check_codestyle.sh
```diff
@@ -0,0 +1,28 @@
...
+    echo "FILES: ${FILES}"
+    clangformat "${FILES}"
+    echo "INCORRECT_FILES: ${INCORRECT_FILES}"
+    [ "$INCORRECT_FILES" ] && exit 1 || exit 0
```
junghans: use `[[.. ]]` and no quotes.

docker/Dockerfile
```diff
@@ -25,6 +26,7 @@ RUN chown -R espressopp:espressopp /home/espressopp/espressopp /home/espressopp/
 USER espressopp
 
 WORKDIR /home/espressopp/espressopp
+RUN maintainer/check_codestyle.sh ${CODESTYLE} ${TRAVIS_BRANCH}
 RUN mkdir build
 
 WORKDIR build
```
junghans: We could put the block below in a `if [[ $CODESTYLE != ON ]]` to make it fails faster, or at least of that for `make` and `make test`.

docker/Dockerfile
```diff
@@ -25,23 +27,29 @@ RUN chown -R espressopp:espressopp /home/espressopp/espressopp /home/espressopp/
...
+        if [ ${CC} = clang ]; then \
+          $HOME/.local/bin/codecov --gcov-exec "llvm-cov gcov"; \
+        else \
+          $HOME/.local/bin/codecov; \
+        fi; \
+      fi; \
+  fi && cd -
 USER root
-RUN make install
+RUN [ "${CODESTYLE}" = ON ] && exit 0 || make install
```
MrTheodor: Why we run again `make install`, this time as a root?

junghans: The first install tests if `DESTDIR` is honored!

docker/Dockerfile
```diff
@@ -25,23 +27,30 @@ RUN chown -R espressopp:espressopp /home/espressopp/espressopp /home/espressopp/
 USER espressopp
 
 WORKDIR /home/espressopp/espressopp
-RUN mkdir build
+RUN maintainer/check_codestyle.sh ${CODESTYLE} ${TRAVIS_BRANCH} ${TRAVIS_PULL_REQUEST} ${TRAVIS_COMMIT} ${TRAVIS_COMMIT_RANGE}
```
junghans: Could we do the `${CODESTYLE}" = ON ` check outside the script, that might make it more clear. Plus we don't have to hand `${CODESTYLE}` inside the script.

docker/Dockerfile
junghans:
```suggestion
RUN [ "${CODESTYLE}" = ON ] || make install
```

docker/Dockerfile
```diff
@@ -25,23 +27,30 @@ RUN chown -R espressopp:espressopp /home/espressopp/espressopp /home/espressopp/
...
-RUN ccache -z
-RUN cmake -DCMAKE_INSTALL_PREFIX=/usr -DEXTERNAL_BOOST=$EXTERNAL -DEXTERNAL_MPI4PY=$EXTERNAL -DWITH_XTC=ON ${COVERAGE:+-DUSE_GCOV=ON} ..
-RUN make VERBOSE=1 -j2 all ${BUILD_UG:+ug doc ug-pdf}
-RUN ccache -s
-RUN make test CTEST_OUTPUT_ON_FAILURE=1
-RUN make install DESTDIR=${PWD}
-RUN cd .. && if [ ${COVERAGE} ]; then \
-  if [ ${CC} = clang ]; then \
-    $HOME/.local/bin/codecov --gcov-exec "llvm-cov gcov"; \
+RUN if [ "${CODESTYLE}" = ON ]; then \
```
junghans:
```suggestion
RUN if [ "${CODESTYLE}" != ON ]; then \
```
and remove the `else` part!

maintainer/check_codestyle.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+echo "check_codestyle.sh CODESTYLE=$1 TRAVIS_BRANCH=$2 TRAVIS_PULL_REQUEST=$3 TRAVIS_COMMIT=$4 TRAVIS_COMMIT_RANGE=$5"
+
+CODESTYLE=$1
+TRAVIS_BRANCH=$2
+TRAVIS_PULL_REQUEST=$3
+TRAVIS_COMMIT=$4
+TRAVIS_COMMIT_RANGE=$5
```
junghans: I think the script inherit the environment, as `ARG` will export environment variables, so I guess we don't need the arguments!

junghans:
@jkrajniak could you do something like this:
https://github.com/laristra/flecsi/commit/7d78b32732e2626d2d4c6bc6c43f310f5f9e990d

Also should be run `pep8` on the python files/

XzzX:
Is there an agreement on a common style?

I personally do not like the two spaces indent. I think four is much more readable.

XzzX:
I suggest:
```
BasedOnStyle: Google
AccessModifierOffset: '-4'
BinPackParameters: 'false'
BreakBeforeBraces: Allman
IndentWidth: '4'
```

junghans:
@XzzX go ahead.

I would abandon this PR though.

PR #208, opened 2018-01-29
.travis.yml: use travis_retry

junghans:
This worksaround network issues when fetching docker image


govarguz:
LGTM!

niktre:
@junghans So, this prevents travis from freezing and making error compilations?

junghans:
@niktre, it will basically retry to fetch the docker image 3 times.

PR #209, opened 2018-02-01
Particle distribution to CPU in MeanSquareInternalDist

hidekb:
The class MeanSquareInternalDist calculate the distance between particles belonging to a same polymer. Hence, particles should be distributed to CPU for each polymer.
I implemented this modification to MeanSquareInternalDist.hpp


govarguz:
@hidekb In order to do so you identify particles belonging to every polymer get local particles per Processor (CPU) and then unfold them so that you count only local particles?

hidekb:
@govarguz In MeanSquareInternalDist (ConfigsParticleDecomp), particles properties are stored into a data structure "configurations". The MeanSquareInternalDist class calculates the physical quantities using the particle configurations in the data structure "configurations". I think that there are no differences between real and local particles in the data structure "configurations".

govarguz:
@hidekb I see, LGTM

niktre:
@hidekb a very small test in the `testsuite/` would be great. It would also increase the coverage and thereby allow automatic merging of this PR.

hidekb:
@niktre Thank you for your suggestion. I added test python script in the testuite/.

govarguz:
LVGTM

PR #214, opened 2018-03-13
Fix assignment of the lattice nodes to CPUs if their ratio is not integer

niktre:
Here I fix a bug resulting in a segmentation fault. The bug appears only if the number of lattice nodes in a specific direction divided by the number of CPUs in the same direction is not integer. 

For example, 20 nodes are divided between 3 CPUs resulting in 6, 7, 7 lattice nodes per CPU. Each CPU also have one "ghost" node in every direction for communication. In MD domain decomposition the boundaries are located at 6.33 and 13.66. Imagine now, that the first CPU has an MD particle at the position, 6.25. To couple this particle to the lattice Boltzmann fluid one needs an info on the LB fluid from the nodes surrounding the particle. These are nodes number 7 and 8 located at 6.0 and 7.0 correspondingly. In the first CPU there is an info on 6 LB nodes (at coordinates 0., 1., 2., 3., 4., 5.) and two "ghost" nodes (one on the left and one on right side with coordinates -1, and 6.). So the node located at 6.0 is covered, but the one at 7.0 is not. This results in a seg.fault. The ceil function in LB domain decomposition remedies the problem.


MrTheodor:
Could we have the test for such bug?

niktre:
@MrTheodor good point. There is an LB/MD coupling test. But the outcome of this test depends on the number of CPUs it is running on. Is there a way to set this number?

MrTheodor:
@niktre It seems to be possible to run on travis tests that use more than one process. See here: https://d-meiser.github.io/2016/01/10/mpi-travis.html
I think the best way is to change CMakeList.txt for the specific testsuite to include mpirun in ```add_test``` command.

MrTheodor:
@niktre please see the example of the tests, they are failing for different number of processes.

niktre:
@MrTheodor in that particular case the tests are expected to fail as the system size is too small (N_i = 5). So it makes no sense to run it even with 2 CPUs. I could increase the system size and test for several number of CPUs, but these test will take longer than the test with a single CPU. Do you think it makes sense?

MrTheodor:
@niktre that depends how sensitive is the code wrt to the number of CPUs and if there is some dependencies?  Here, you could e.g. create a unit test that will only cover the lattice creation and run it for different number of cpus.

niktre:
@MrTheodor  do you have any idea why the test for 1 CPU is failing? Without `foreach(PROCS ...)` it was working

testsuite/lattice_boltzmann/test_LatticeBoltzmann.py
```diff
@@ -27,18 +27,19 @@
...
 initDen = 1.
 initVel = 0.
 initVelSin = 0.1
 
 class TestPureLB(unittest.TestCase):
     def setUp(self):
         # set up system
-        global Ni, temperature
-        system, integrator = espressopp.standard_system.LennardJones(0, box=(Ni, Ni, Ni), temperature=temperature)
+        global temperature
```
MrTheodor: why do you need those global variables?  you could set them in the `setUp` method and then use `self.temperature`, `self.runSteps`

MrTheodor:
@niktre what was the reason? precision or run time?

niktre:
@MrTheodor precision was broken for two narrow directions of the box

testsuite/lattice_boltzmann/test_LatticeBoltzmann.py
niktre: thanks, I've corrected this. But this code was partially overwritten in the last commit

PR #215, opened 2018-04-16
Test the build against the latest ubuntu

MrTheodor:


junghans:
Should we deploy this, too? If so, we need to make the `DOCKER_TAG` unique for each distribution.

MrTheodor:
I am not sure we should complicate with the distro versions for the images. In the end, "user" (if any) would like simply use an image and do not care about the versions. Is Dockery popular in HPCs? 
For the deploy, I would use the latest stable version and not LTS.

junghans:
Ok, currently only `DISTRO=ubuntu` gets deployed, so let's leave that like it is for now.

PR #231, opened 2018-07-17
cmake: allow user to override PYTHON_INSTDIR

junghans:


jkrajniak:
What is this variable for?

junghans:
To customize the installation location of the python package!

PR #241, opened 2018-08-10
Smooth square well

songbin6280:
Hello,

I have implemented a smooth sqaure-well potential introduced in the paper by Leitold and Dellago in JCP 2014. It is implemented based on VerletListInteractionTemplate, FixedPairListInteractionTemplate and FixedPairListTypesInteractionTemplate. I have been running simulations with this potential for a while. I would like to distribute the code with espresso++. Please take a look at it.

Best,
Bin


src/interaction/SmoothSquareWell.py
```diff
@@ -0,0 +1,101 @@
...
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+r"""
+***************************************
+espressopp.integrator.SmoothSquareWell
+***************************************
+This is an inplementation of the smoothed square-well potential from Leitold and Dellago JCP 141 (2014).
```
jkrajniak: Please, put here the equation that this potential realized. Also, would be good to have url to the article instead of a weak citation.

e.g.

```
.. math::

   \frac{ \sum_{t=0}^{N}f(t,k) }{N}
```

src/interaction/SmoothSquareWell.py
```diff
@@ -0,0 +1,101 @@
...
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+r"""
+***************************************
+espressopp.integrator.SmoothSquareWell
```
jkrajniak: should be `espressopp.interaction.SmoothSquareWell`

src/interaction/SmoothSquareWell.py
```diff
@@ -0,0 +1,101 @@
...
+
+from espressopp.interaction.Potential import *
+from espressopp.interaction.Interaction import *
+from _espressopp import interaction_SmoothSquareWell, interaction_VerletListSmoothSquareWell, \
+    interaction_FixedPairListSmoothSquareWell
+
+class SmoothSquareWellLocal(PotentialLocal, interaction_SmoothSquareWell):
+
+    def __init__(self, epsilon=1.0, sigma=0.0,
+                 cutoff=infinity, shift=0.0):
```
jkrajniak: no need to have new line

songbin6280:
I will make changes and create tests. Thank you.

songbin6280:
Hello,

I have a question about the unit test. I wrote the unit test with the Boost unit test framework, as some of the tests in the unittest directory in the interaction directory. Then I noticed make test seems not to be testing on these files. Are you only accepting those tests done in the testssuite with a python script?

Bin

jkrajniak:
@songbin6280 for boost tests, please have a look here: https://github.com/espressopp/espressopp/tree/master/testsuite/Array4DTest

However, I think what you have to test is the methods `computeEnergy(r)` and all the setters and getters. E.g. what is the expected value of energy when epsilon is 0, etc.
This can be done very easily at the level of Python tests. In addition, this so-called integration tests will check the correctness of Python interface.

src/interaction/SmoothSquareWell.cpp
```diff
@@ -41,7 +42,7 @@ namespace espressopp {
         .def(init< real, real, real, real >())
         .add_property("epsilon", &SmoothSquareWell::getEpsilon, &SmoothSquareWell::setEpsilon)
         .add_property("sigma", &SmoothSquareWell::getSigma, &SmoothSquareWell::setSigma)
-        .add_property("width", &SmoothSquareWell::getLambda, &SmoothSquareWell::setLambda)
+        .add_property("Lambda", &SmoothSquareWell::getLambda, &SmoothSquareWell::setLambda)
```
jkrajniak: make it consistent, so `lambda`

src/interaction/SmoothSquareWell.py
```diff
@@ -83,7 +99,7 @@ class SmoothSquareWell(Potential):
         'The SmoothSquareWell potential.'
         pmiproxydefs = dict(
             cls = 'espressopp.interaction.SmoothSquareWellLocal',
-            pmiproperty = ['epsilon', 'sigma', 'width', 'a']
+            pmiproperty = ['epsilon', 'sigma', 'Lambda', 'a']
```
jkrajniak: `lambda`

songbin6280: lambda wouldn't work because it is a python keyword, unfortunately.

jkrajniak: True, forgot about that, soo keep like it is

jkrajniak:
Yeah, sorry; I will merge as soon as build pass.

songbin6280:
No problem. Thank you for the quick reply.

PR #249, opened 2018-12-18
Change to the comment on the HeSpaDDA

govarguz:
@stuehn, as agreed I am modifying the comment on the HeSpaDDA algorithm and the old we had.


govarguz:
@niktre smth else I could try?

PR #256, opened 2019-02-01
Updated messages and added source file for LOG4ESPP_LEVEL_TRACE to work

jnvance:


src/iterator/CellListAllTriplesIterator.cpp
```diff
@@ -0,0 +1,29 @@
+/*
+  Copyright (C) 2012,2013
```
junghans: 2019?

jkrajniak: Not sure, copyright should rather represent full history of the contributors (a.k.a. copyright owners).

PR #244, opened 2018-08-25
Move all unittests to testsuite/

jkrajniak:
1. Migrate all tests spread along `src/*/unittest` to `testsuit/`. As they weren't run in a continuous fashion, mostly they are outdated. I've tried to bring the life into them again but in the end, I am not sure it makes sense. 
2. Updated `testsuite/CMakeList.txt` to import the tests automatically whenever the `CMakeList.txt` file is not present in the particular test directory.

PR #258, opened 2019-02-01
Version bump to 2.0.1

junghans:
With all the bug fixes, we should make another release.

Maybe merge #257, #256, #255, #244 first.


junghans:
@jkrajniak merge when ready!

PR #260, opened 2019-02-02
Coverage calculation only on one step

### LINKED ISSUE #148: Drop some coverage builds
I seems ccache doesn't play well with coverage and hence these builds take 26min instead of the usual 7min.

We are currently doing 8 builds (2 compiler * 2 mpi variants * 2 distributions), but 2 builds might be enough as mpi and distribution won't change the coverage rate anyhow.

jkrajniak:
Apparently, the coverage calculation slows down the entire process. Without coverage, a single step builds in around 5-8 min, with coverage this is something like 41 min.

Fix #148


docker/Dockerfile
```diff
@@ -17,6 +17,8 @@ ARG TRAVIS_TAG
 ARG TRAVIS_REPO_SLUG
 ARG TRAVIS_COMMIT
 
+ENV TEST_H5MD ON
```
junghans: Should we make it an `ARG`?

jkrajniak: I would simply keep it switched `ON` and test H5MD all the time.

junghans: I guess then I wouldn’t make a environment variable at all!

PR #219, opened 2018-05-15
Dockerfile: filter gcov arcs warnings

junghans:
https://bugzilla.redhat.com/show_bug.cgi?id=1577508

PR #263, opened 2019-02-04
.codecov.yml: exclude /usr/include

junghans:
Good part of #219


jkrajniak:
maybe also `contrib/` ?

junghans:
Good point!

PR #269, opened 2019-02-28
.travis.yml: fix deploy

junghans:


junghans:
This should fix the deploy to https://hub.docker.com/r/espressopp/espressopp

PR #273, opened 2019-03-11
doc: add reference to 2019 esp++ paper

junghans:


junghans:
@hidekb good catch - fixed!

PR #278, opened 2019-03-18
bump cmake required version to 3.11 and requires boost-python27

### LINKED ISSUE #277: Could not find the following Boost libraries: boost_python with boost-1.69 and cmake-3.14
With cmake-3.14 and python-1.69, I got:
```
CMake Error at /opt/cmake-3.14.0/share/cmake-3.14/Modules/FindBoost.cmake:2132 (message):
  Unable to find the requested Boost libraries.

  Boost version: 1.69.0

  Boost include path:
  /opt//boost-1.69.0/include


  Could not find the following Boost libraries:

          boost_python

  Some (but not all) of the required Boost libraries were found.  You may
  need to install these additional Boost libraries.  Alternatively, set
  BOOST_LIBRARYDIR to the directory containing Boost libraries or BOOST_ROOT
  to the location of Boost.
```

The core of the issue can be found in `FindBoost.cmake`:

> Note that Boost Python components require a Python version suffix
> (Boost 1.67 and later), e.g. ``python36`` or ``python27`` for the
> versions built against Python 3.6 and 2.7, respectively.

This basically means for Boost-1.67 and later, we need to check for the `python27` component instead of just `python`, however this feature is only supported in cmake-3.11 and newer.

jkrajniak:
- required python27 from boost library

fix #277


CMakeLists.txt
```diff
@@ -152,7 +152,7 @@ list(APPEND EXTRA_LIBRARIES ${CMAKE_DL_LIBS})
 
 if (EXTERNAL_BOOST)
   set(NEEDED_Boost_VERSION 1.58.0)
-  find_package(Boost ${NEEDED_Boost_VERSION} REQUIRED COMPONENTS mpi python serialization system filesystem)
+  find_package(Boost ${NEEDED_Boost_VERSION} REQUIRED COMPONENTS mpi python27 serialization system filesystem)
```
junghans:
```suggestion
  find_package(Boost ${NEEDED_Boost_VERSION} REQUIRED COMPONENTS mpi serialization system filesystem)
  if(Boost_VERSION VERSION_LESS 1.67)
     find_package(Boost ${NEEDED_Boost_VERSION} REQUIRED COMPONENTS python)
   else()
     find_package(Boost ${NEEDED_Boost_VERSION} REQUIRED COMPONENTS python27)
   endif()
 endif()
```

jkrajniak:
```
Traceback (most recent call last):
  File "/home/espressopp/espressopp/testsuite/AdResS/RDFatomistic/test_RDFatomistic.py", line 23, in <module>
    import espressopp
  File "/home/espressopp/espressopp/build/espressopp/__init__.py", line 23, in <module>
    from espressopp.main._setup import *
  File "/home/espressopp/espressopp/build/espressopp/main/_setup.py", line 26, in <module>
    import _espressopp
ImportError: /home/espressopp/espressopp/build/_espressopp.so: undefined symbol: _ZTIN5boost7archive6detail14basic_iarchiveE
```
completely confused

junghans:
I think the problem is that we are using `Boost_LIBRARIES`, which get overwritten by the second `find_package(Boost)` call.

niktre:
@jkrajniak LGTM now

docker/Dockerfile.local
```diff
@@ -14,6 +14,6 @@ RUN mkdir build
 
 WORKDIR build
 RUN cmake -DCMAKE_INSTALL_PREFIX=/usr -DEXTERNAL_BOOST=$EXTERNAL -DEXTERNAL_MPI4PY=$EXTERNAL -DWITH_XTC=ON ..
-RUN make -j4 all
-RUN make install
-WORKDIR ../examples
+#RUN make -j4 all
+#RUN make install
+#WORKDIR ../examples
```
junghans: Why?

niktre: maybe to save time and see that building is working?

junghans:
What about the boost detection in testsuite?

jkrajniak:
@junghans  Can this be shared between CMakeList ?

junghans:
Yeah, we could share the variable!

PR #280, opened 2019-03-19
Halfcell

ppkk:
basic implementation of the halfcell algorithm


src/storage/DomainDecomposition.py
```diff
@@ -54,9 +54,9 @@
```
jkrajniak: Please update the documentation section with the information about the new parameter.

govarguz:
@ppkk LGTM with the following minor revs:

1. As  @jkrajniak pointed out, we need a bit of documentation, and a small warning can be added whenever a system with "small cutoff " (e.g. less than 2) and the user decides to use a half or third cell which might introduce an overhead (for sure).
2. you can upate the copyright year to the files you modified 2019
3.(optional) perhaps you would like to add the MPCDF to the copyright?

niktre:
@ppkk @govarguz Agree with Horacio on 1, 2 and 3. Though on point 1 I would not call it "small", as it is quite relative. Just put is as "if r_cut < 2 you may not see significant improvements with respect to full cells. Moreover, in some cases there can be an insignificant performance drop".

jkrajniak:
@ppkk any update with this PR?

ppkk:
I documented the halfCellInt parameter and updated the copyright info. I am not sure with the warning yet. is the r_cut value enough to asses the performance? Does not the particle density also play a role?

govarguz:
well, In fact this depends on the r_cut and hence the Cell Size itself where the particles will be iterated. And you know the number of particles per cell (once you decompose the sys. This will be one parameter that can be looked at). However, as you checked even with very dense 0.85 systems for smaller cutoffs (e.g. ~1.12) and hence Cell-sizes, no performance improvement was noticeable. Anyhow, feel free to propose another way to show the warning. Best,

PR #283, opened 2019-04-19
add ability for RNG to restart from a prior state

songbin6280:
I have added two functions to the RNG class so that it is possible to save the state RNG to a file and restart from the same state from the file.


songbin6280:
Hello Jakub,

I have made requested changes to the code.

Best,
Bin

PR #292, opened 2020-01-04
Speedup verlet list rebuild by copying neighbors into buffers

jnvance:
During each Verlet list rebuild, particle positions of neighboring cells of every cell are first copied into contiguous buffers (`c_pos`). The buffers are then iterated over to determine the particle pairs that are within cutoff distance. This results in more cache-friendly access since the particle positions are stored contiguously instead of being stored in different memory locations (in the case of `particles` in `std::vector< Cell >`) and separated by other particle information. The cost of copying the data initially is amortized by the number of times the buffer is iterated during the cutoff check (equal to the number of particles in the cell).

I've added a test and comparison script that runs a Lennard-Jones simulation (`testsuite/verlet_list_buffer/verlet_list_buffer.py`). With one process on a Broadwell node of Mogon 2, compiled with gcc 7.3.0 and -O3 optimization, the method results in around 2x speedup for the *Resort* part of the simulation.


src/VerletList.cpp
```diff
@@ -84,25 +86,180 @@ namespace espressopp {
...
     
     vlPairs.clear();
 
-    // add particles to adress zone
-    CellList cl = getSystem()->storage->getRealCells();
-    LOG4ESPP_DEBUG(theLogger, "local cell list size = " << cl.size());
-    for (CellListAllPairsIterator it(cl); it.isValid(); ++it) {
-      checkPair(*it->first, *it->second);
-      LOG4ESPP_DEBUG(theLogger, "checking particles " << it->first->id() << " and " << it->second->id());
+    if(useBuffers) {
```
jkrajniak: Could you check if this `if (useBuffers)` has some impact on performance in case of `useBuffers = true|false` and if you modify the code by removing if and keeping only `rebuildUsingBuffers` statement.

src/VerletList.cpp
```diff
@@ -84,25 +86,180 @@ namespace espressopp {
...
+      rebuildUsingBuffers(exList.size(), useSOA);
     }
```
jkrajniak: Please, keep the formatting consistent
` } else {`

src/VerletList.cpp
```diff
@@ -84,25 +86,180 @@ namespace espressopp {
...
+    const CellList& realCells = getSystem()->storage->getRealCells();
+    const size_t numRealCells = realCells.size();
+
+    // stores the range of neighbor particles belonging to cell i: c_range[i] to c_range[i+1]
+    std::vector<int> c_range;
+    c_range.reserve(numRealCells+1);
+
+    // get the number of particles in all neighbor cells
+    size_t c_reserve = 0;
+    c_range.push_back(c_reserve);
```
jkrajniak: Why do you have to push `0` as the first element?

src/VerletList.cpp
```diff
@@ -84,25 +86,180 @@ namespace espressopp {
...
+      if(USE_EXCLUSION_LIST){
+        c_id.resize(c_resize);
+      }
+    }
+
+    // fill buffer
+    size_t ip = 0;
+    for(size_t icell=0; icell<numRealCells; icell++)
+    {
+      // size_t ip = c_range[icell];
```
jkrajniak: Please, remove  commented code

src/VerletList.cpp
```diff
@@ -84,25 +86,180 @@ namespace espressopp {
...
+      size_t end = c_range[icell+1];
+      for(NeighborCellInfo& nc: realCells[icell]->neighborCells)
+      {
+        if(!nc.useForAllPairs)
+        {
```
jkrajniak: Formatting

src/VerletList.py
```diff
@@ -25,14 +25,18 @@
 *********************
 
 
-.. function:: espressopp.VerletList(system, cutoff, exclusionlist)
+.. function:: espressopp.VerletList(system, cutoff, exclusionlist, useBuffers, useSOA)
 
 		:param system: 
 		:param cutoff: 
 		:param exclusionlist: (default: [])
+		:param useBuffers: Whether particle neighbors are buffered to improve rebuild times. (default: False)
```
jkrajniak: :thinking: in the constructor `useBuffers = true` and here `(default: False)`

src/VerletList.hpp
```diff
@@ -85,11 +87,41 @@ namespace espressopp {
...
   protected:
 
+    std::vector<real> c_x,c_y,c_z;
+    std::vector<Real3D> c_pos;
+    std::vector<Particle*> c_p;
+    std::vector<size_t> c_id, c_type;
+
+    inline void rebuildUsingBuffers(bool useExList, bool useSOA)
+    {
+      if(useExList) {
```
jkrajniak: Does the template here speed up?

src/VerletList.cpp
jnvance: Edited.

src/VerletList.cpp
jnvance: Removed comment.

src/VerletList.cpp
jnvance: Edited formatting for consistency.

src/VerletList.py
jnvance: Sorry, they were all initially False. Now I've set everything to True by default including in the C++ class.

src/VerletList.cpp
jnvance: This initially stored the range of neighbor particles belonging to icell, `size_t start=c_range[icell], end=c_range[icell+1];`. I've now changed it to `size_t end=c_range[icell];` and removed the need for 0 as first element of `c_range`.

src/VerletList.cpp
jnvance: This does not impact performance since the branch is only reached once every rebuild. It doesn't take a lot of time compared to the more critical loops under it. I also checked with VTune and the branch is not marked as a hotspot.

src/VerletList.hpp
jnvance: Yes though I'm not sure by how much. The template allows the compiler to skip compilation of code under `if(USE_EXCLUSION_LIST)` when the parameter is set to 0. I can confirm this with the source code / assembly viewer on VTune which shows that these source code sections do not have corresponding entries in the assembly.

PR #294, opened 2020-01-08
Vectorization

jnvance:
This PR adds a preliminary version of the `vectorization` submodule which speeds up the calculation of short-range pair interactions targeting Intel CPUs that support AVX2 and AVX-512.

When the `espressopp.vectorization.Vectorization` class is instantiated, it connects some of its methods to the integrator and storage objects (currently, only VelocityVerlet and DomainDecomposition are supported).

During every time step of `integrator.run()`, the `Vectorization` class copies the necessary data from the `Particle` lists in the cells (currently only position) into `ParticleArray`, performs the force calculation, and then copies back the forces to the `Particle` lists. The data in `ParticleArray` are stored in a structure of arrays (SOA) layout.  This enables the vectorization of the neighbor-list construction and force calculation using only `#pragma vector` directives (no intrinsics), similar to how it is done in the LAMMPS-INTEL package.

Most of the additional source code is contained in the directory `src/vectorization`, except for two new Boost signals:

- onCellAdjust - called when the cell structure of the storage class is modified
- aftCalcFLocal - called from `integrator.run` after calcForces and just before exchanging ghost forces with neighboring subdomains

Vectorized versions of the VerletList and LennardJones classes are implemented in the vectorization submodule. So to use these on current scripts, one only needs to add the `vectorization` keyword and pass the correct arguments, e.g.

```
vec         = espressopp.vectorization.Vectorization(system, integrator)
verletlist  = espressopp.vectorization.VerletList(system, vec, r_cutoff)
potential   = espressopp.vectorization.interaction.LennardJones(...)
interaction = espressopp.vectorization.interaction.VerletListLennardJones(verletlist)
```

In this version, the energy and virial calculations still rely on the previous VerletList layout, so the Verlet list rebuild has to be manually triggered before calling any analysis functions, e.g.

```
verletlist.rebuildPairs()
espressopp.tools.analyse.info(system, integrator)
```

An example script is provided in `examples/vectorization/lennard_jones.py` which shows the minimal modifications needed for the `examples/lennard_jones/lennard_jones.py` script to use vectorized versions of the routines (currently only for the equilibration  phase since LennardJonesCapped is not yet implemented).

A test was also added in `testsuite/vectorization` to verify that particle trajectories resulting from the MD integration are almost equal whether the vectorized or the non-vectorized version is used.

Development of the submodule is optimized for the Intel 2018 compiler with `-O3 -xHost -restrict` flags for Broadwell, and `-O3 -xSKYLAKE-AVX512 -qopt-zmm-usage=high` for Skylake. Improvements can also be observed with GCC as long as the appropriate optimization flags are used.


junghans:
How does this approach compares to https://github.com/ECP-copa/Cabana?

jnvance:
> How does this approach compares to https://github.com/ECP-copa/Cabana?

Cabana seems to provide a more generalized way of representing Struct-of-Arrays and Array-of-Struct-of-Arrays (where attributes are template arguments) similar to how the data are stored in `ParticleArray` (where the attributes are explicitly listed). Our approach, on the other hand, still uses the original `std::vector<Particle>` (an extended AOSOA) to store particle information everywhere else but just copies necessary information, like position, to the more compact and aligned SOA or AOSOA layout in ParticleArray.

docker/Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN mkdir build
 
 WORKDIR build
 RUN ccache -z
-RUN cmake -DCMAKE_INSTALL_PREFIX=/usr -DEXTERNAL_BOOST=$EXTERNAL -DEXTERNAL_MPI4PY=$EXTERNAL -DWITH_XTC=ON ${COVERAGE:+-DUSE_GCOV=ON} ..
+RUN cmake -DCMAKE_INSTALL_PREFIX=/usr -DEXTERNAL_BOOST=$EXTERNAL -DEXTERNAL_MPI4PY=$EXTERNAL -DWITH_XTC=ON ${COVERAGE:+-DUSE_GCOV=ON} -DCMAKE_CXX_FLAGS="-Wno-deprecated-declarations" ..
```
junghans: This introduces technical debt, the warnings should really be fixed instead.

jnvance: Okay, will revert the commit.

junghans:
> > How does this approach compares to https://github.com/ECP-copa/Cabana?
> 
> Cabana seems to provide a more generalized way of representing Struct-of-Arrays and Array-of-Struct-of-Arrays (where attributes are template arguments) similar to how the data are stored in `ParticleArray` (where the attributes are explicitly listed). Our approach, on the other hand, still uses the original `std::vector<Particle>` (an extended AOSOA) to store particle information everywhere else but just copies necessary information, like position, to the more compact and aligned SOA or AOSOA layout in ParticleArray.

Cabana has support for GPU data layouts, too.

src/vectorization/Vectorization.py
```diff
@@ -0,0 +1,58 @@
...
+    :param integrator: integrator object
+    :param mode: (default='' equiv to 'SOA') 'SOA' for structure of arrays and 'AOS' for array of structures
+
+"""
+
+class VectorizationLocal(_espressopp.Vectorization):
+
+    def __init__(self, system, integrator, mode=""):
+        if pmi.workerIsActive():
+            if mode=="":
```
jkrajniak: This whole `if .... else ... if ... ` could be simplified. 
```python
if `mode == "" or mode == AOS:
    mode_int = _espressopp.VectorizationMode.AOS
elif mode == SOA:
    mode_int = _espressopp.VectorazationMode.SOA
else:
    raise ValueError('Incorrect mode [{}]'.format(mode))
```

src/vectorization/VerletList.cpp
```diff
@@ -0,0 +1,1047 @@
...
+                  else
+                  {
+                    dist_x   = p_x - pa_p_x[np];
+                    dist_y   = p_y - pa_p_y[np];
+                    dist_z   = p_z - pa_p_z[np];
+                  }
+
+                  const real distSqr  = dist_x*dist_x + dist_y*dist_y + dist_z*dist_z;
+
+                  if(p>=np) include_np = 0;
```
jkrajniak: I suppose these three ifs can be simplified to 
```cpp
if (p < np && distSqr <= cutsq) {
  npptr[ll++] = np;
}

```

src/vectorization/VerletList.cpp
```diff
@@ -0,0 +1,1047 @@
...
+                  else
+                  {
+                    dist_x   = p_x - pa_p_x[np];
+                    dist_y   = p_y - pa_p_y[np];
+                    dist_z   = p_z - pa_p_z[np];
+                  }
+
+                  const real distSqr  = dist_x*dist_x + dist_y*dist_y + dist_z*dist_z;
+
+                  if(distSqr > cutsq) include_np = 0;
```
jkrajniak: Here as well
```cpp
if (distSqr <= cutsq) {
  npptr[ll++] = np;
}
```

src/vectorization/VerletList.cpp
```diff
@@ -0,0 +1,1047 @@
...
+                  else
+                  {
+                    dist_x   = p_x - pa_p_x[np];
+                    dist_y   = p_y - pa_p_y[np];
+                    dist_z   = p_z - pa_p_z[np];
+                  }
+
+                  const real distSqr  = dist_x*dist_x + dist_y*dist_y + dist_z*dist_z;
+
+                  if(p>=np) include_np = 0;
```
jkrajniak: perhaps the same simplification as above

src/vectorization/ParticleArray.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+#include "ParticleArray.hpp"
+#include "Cell.hpp"
+#include <iostream>
+
+///////////////////////////////////////////////////////////////////////////////////////////////////
+namespace espressopp { namespace vectorization
+{
+  /////////////////////////////////////////////////////////////////////////////////////////////////
+  void ParticleArray::copy_from(CellList const& srcCells, Mode mode_)
```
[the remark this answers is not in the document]
jnvance: Renamed methods to camel case.

src/vectorization/Vectorization.py
```diff
@@ -0,0 +1,58 @@
...
+**************************************
+
+.. function:: espressopp.vectorization.Vectorization(system, integrator, mode)
+
+    :param system: system object
+    :param integrator: integrator object
+    :param mode: (default='' equiv to 'SOA') 'SOA' for structure of arrays and 'AOS' for array of structures
+
+"""
+
```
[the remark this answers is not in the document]
jnvance: Added labels.

src/vectorization/Vectorization.py
jnvance: Simplified block.

src/vectorization/VerletList.cpp
jnvance: Simplified if block

src/vectorization/VerletList.cpp
jnvance: Simplified if block

src/vectorization/VerletList.cpp
jnvance: Simplified if block

src/vectorization/Vectorization.py
```diff
@@ -34,20 +34,20 @@
...
 class VectorizationLocal(_espressopp.Vectorization):
 
     def __init__(self, system, integrator, mode=""):
         if pmi.workerIsActive():
-            if mode=="":
-                cxxinit(self, _espressopp.Vectorization, system, integrator)
+            if mode=="" or mode==SOA:
+                mode_int = _espressopp.VectorizationMode.SOA
+            elif (mode==AOS):
```
jkrajniak: You don't need parentheses

jnvance: Removed parentheses

jnvance:
It seems the CI failures all result from `The job exceeded the maximum log length, and has been terminated.` during the `make install` phase of the test script.  Most of the log is from deprecation warnings about std::auto_ptr from the internal Boost library on Ubuntu with GCC. The limit is around 4MB, and even the corresponding logs for the master branch are close to exceeding this limit.

What do you guys think would be the best way to remove or suppress the warnings?

jkrajniak:
Let's try with `-Wdeprecated-declarations`

jnvance:
> Let's try with `-Wdeprecated-declarations`

I placed that flag in the [Dockerfile](https://github.com/espressopp/espressopp/pull/294/commits/374b8244350b3cd63a22baf0979140674923e5d4) before but reverted it on advice of @junghans 

Can we also place that flag directly in `.travis.yml` just for `env: DISTRO=ubuntu EXTERNAL=OFF`, `DISTRO=ubuntu_rolling EXTERNAL=OFF` and `DISTRO=ubuntu_mpich EXTERNAL=OFF` with `compiler: gcc`?

jkrajniak:
> > Let's try with `-Wdeprecated-declarations`
> 
> I placed that flag in the [Dockerfile](https://github.com/espressopp/espressopp/pull/294/commits/374b8244350b3cd63a22baf0979140674923e5d4) before but reverted it on advice of @junghans
> 
> Can we also place that flag directly in `.travis.yml` just for `env: DISTRO=ubuntu EXTERNAL=OFF`, `DISTRO=ubuntu_rolling EXTERNAL=OFF` and `DISTRO=ubuntu_mpich EXTERNAL=OFF` with `compiler: gcc`?

You can try; These warnings are related to boost library, I wouldn't care too much about switching this off.

junghans:
@jnvance see https://github.com/votca/tools/blob/master/include/votca/tools/eigen.h#L34-L55 for an example on how to disable warning from an included header.

src/include/python.hpp
```diff
@@ -1,4 +1,6 @@
 /*
+  Copyright (C) 2018-2019
```
jkrajniak: @junghans do you think this copyright of VOTCA team should be placed here?

PR #295, opened 2020-02-06
Fix failing gitlab build

junghans:


junghans:
`polymer_melt_tabulated_thirdcell` times out.

junghans:
@jnvance @jkrajniak can you have a look?

src/VerletList.cpp
```diff
@@ -232,19 +232,16 @@ namespace espressopp {
             distsq = d.sqr();
           }
 
-          bool addpair = true;
-          if (distsq > cutsq) addpair = false;
+          if (distsq > cutsq) continue;
 
           if(USE_EXCLUSION_LIST) {
```
jkrajniak: How about to move this exclusion list check at the begining of `for` loop (let say line 224) ?

jnvance: The exclusion list check seems to be more expensive than the cut-off distance check. For example with `testsuite/pi_water/water.py`, the rebuild takes 24.94 seconds, while moving the exclusion list check before cut-off check takes 113.49 seconds. (Intel(R) Xeon(R) Gold 6130 CPU @ 2.10GHz, Intel compiler 18.0.3, CXX_FLAGS="-g -O3 -restrict -xHost -DNDEBUG -qopt-zmm-usage=high")

PR #302, opened 2020-04-18
.travis.yml: fix docker deploy

junghans:
See travis-ci/travis-ci#10512 for details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ jobs:
           env: DISTRO=ubuntu EXTERNAL=ON BUILD_UG=ON
 
 after_success:
-  - test "${TRAVIS_BUILD_STAGE_NAME}" = "Deploy" || travis_terminate 0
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -47,7 +47,7 @@ jobs:
           env: DISTRO=ubuntu EXTERNAL=ON BUILD_UG=ON
 
 after_success:
-  - test "${TRAVIS_BUILD_STAGE_NAME}" = "Deploy" || travis_terminate 0
+  - if [[ "${TRAVIS_BUILD_STAGE_NAME}" =! [dD]eploy ]]; then travis_terminate 0; fi
```
jkrajniak, line 50: Shouldn't be !=

junghans: It was 'or' (`||`), now it is 'and' due to the if statement hence I had to negate the condition.

jkrajniak: Just the condition `=!` ? shouldn't be `!=`

junghans: Ah....good point.

junghans: Fixed

PR #304, opened 2020-04-20
FIX: pdbreader according to pdb format, and updated pdb format reference

govarguz:
FIX: pdbreader according to pdb format, and updated pdb format reference to http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM


govarguz:
Seems, to be my former commits where not added! Thanks for having a look

junghans:
```
======================================================================
ERROR: test_pdb_read_write (__main__.TestPythonTools)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/espressopp/espressopp/testsuite/python_tools/io/test_python_tools_io.py", line 77, in test_pdb_read_write
    index,atomname,resname,resid,x,y,z,alpha,beta,segid,element = espressopp.tools.pdbread(filename,natoms=2,header=2)
ValueError: too many values to unpack
----------------------------------------------------------------------
Ran 3 tests in 0.006s
FAILED (errors=1)
```

PR #312, opened 2020-09-22
Update DumpXTC for latest GROMACS version

jkrajniak:
Package libgromacs-devel does not provide `gromacs/fileio/xtcio.h` on ubuntu; This is more general problem as GROMACS 2020 changed the available header files. I have update DumpXTC and DumpXTCAddress to use `gromacs/fileio/trxio.h` and appropriate methods.

cc: @jnvance

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -18,7 +18,7 @@ stages:
     - j="$(grep -c processor /proc/cpuinfo 2>/dev/null)" || j=0; ((j++))
     - mkdir -p build
     - pushd build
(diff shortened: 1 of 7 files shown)


junghans:
Something is wrong in the code, one can write xtc file by using `gromacs/fileio/trxio.h`, which support `xtc` and `trr` files.

/cc @GDeichmann

junghans:
i think the key is to use `open_trx()` and `write_trxframe()` instead of `open_xtc()` and `write_xtc()`.

jkrajniak:
> i think the key is to use `open_trx()` and `write_trxframe()` instead of `open_xtc()` and `write_xtc()`.

Let see, I have made adjustments. I'm wondering which distribution will fail

junghans (approved):

PR #317, opened 2020-10-08
Require compiler to be compatible with C++14

jkrajniak:


junghans (approved):

jkrajniak:
Great, gitlab compilation works

PR #322, opened 2021-01-25
use long long step instead of int

songbin6280:
Dear reviewers,

The type of step is changed into long long, as step from the integrator is used in the argument, which is of long long type.

Bin.


junghans (approved):

XzzX:
Already changed in the master. Thank you for the PR.

PR #323, opened 2021-03-24
Faster adding of particles using a numpy array

jnvance:
Added the function `addParticlesArray` to the Storage class. It has the same signature as `addParticles` but it instead accepts a numpy array as the input for the particle data.

This adds the Boost.python.numpy library as an optional requirement, and skips compilation of the underlying functions if the library is not found.

The speedups can vary greatly depending on the number of particles and procs but for example I managed to get the following results with the example scripts `examples/add_particles_numpy/add_particles_numpy.py` on a single Skylake node with 32 MPI ranks:

```
Using system.storage.addParticles(...)
    prepared 1000000 particles in 4.71 seconds
    added    1000000 particles in 23.76 seconds

Using system.storage.addParticlesArray(...)
    prepared 1000000 particles in 0.56 seconds, speedup: 8.39
    added    1000000 particles in 3.53 seconds, speedup: 6.74
```


junghans:
@jnvance the CI is borked, can you have a look at that too. We might actually also need to migrate from Travis to GitHub actions.

jnvance:
> @jnvance the CI is borked, can you have a look at that too. We might actually also need to migrate from Travis to GitHub actions.

I had a quick look, seems to be a problem with mpi4py and the io modules. I'll check in the docker images later.

src/storage/Storage.cpp
```diff
@@ -743,12 +748,199 @@ namespace espressopp {
...
+
     //////////////////////////////////////////////////
     // REGISTRATION WITH PYTHON
     //////////////////////////////////////////////////
     void
     Storage::registerPython() {
       using namespace espressopp::python;
+
+      #ifdef ESPP_BOOST_HAS_NUMPY
+      numpy::initialize();
```
XzzX, line 941: Should be in `_espresso.cpp` to ensure it gets called regardless of `Storage`.

src/storage/Storage.py
```diff
@@ -428,12 +429,109 @@ def printRealParticles(self):
...
+            if properties == None:
+                # default properties = (id, pos)
+                index_id = 0
+                index_posx = 1
+                index_posy = 2
+                index_posz = 3
+                nindex = 4
+            else:
+                nindex = 0
+                for val in properties:
```
XzzX, line 479:
```suggestion
                for nindex, val in enumerate(properties):
```

examples/add_particles_numpy/add_particles_numpy.py
```diff
@@ -0,0 +1,87 @@
...
+        props = ['id', 'type', 'mass', 'posx', 'posy', 'posz', 'vx', 'vy', 'vz']
+        ids      = np.arange(1,num_particles+1)
+        types    = np.zeros(num_particles)
+        mass     = np.ones(num_particles)
+        x        = np.array(x)
+        y        = np.array(y)
+        z        = np.array(z)
+        vx       = np.array(vx)
+        vy       = np.array(vy)
+        vz       = np.array(vz)
```
XzzX, line 53: We could use two-dimensional arrays to save some parameters. At the same time it ensures that all three components are present.

XzzX (changes requested):
I highly appreciate the addition of `numpy` as it is the de facto standard for scientific computing in `python`. We should make it a requirement and not optional. For the future I think `np.array` should be the standard exchange format between Python and C++.

Merge should wait till the refactoring of cmake is finished and add the boost numpy target.

XzzX:
We should add `numpy` to requirements.txt

jnvance:
This actually requires boost-python-numpy library which is not often installed by default. Is it installed in the systems we use for testing?

XzzX:
> This actually requires boost-python-numpy library which is not often installed by default. Is it installed in the systems we use for testing?

Ubuntu and Fedora yes, OpenSUSE does separate installs of boost libraries. @junghans 
From my perspective I would either install boost completely or not at all. I would not pick individual libraries. Since espressopp already makes use of a lot of boost libraries I think we can add boost-python-numpy to the list. I am open for discussion.

junghans:
Yeah, let's install it! Not sure why openSUSE splits all their libraries.

jnvance:
I agree, the entire boost library should be a dependency. But from my experience, the boost-python-numpy library is built only when numpy is detected in the python installation so this might not always be available for all users (?)

I was thinking we should have a way to handle optional dependencies as well.

junghans:
Can you add the necessary packages in a PR to espressopp/buildenv?

examples/add_particles_numpy/add_particles_numpy.py
jnvance: Problem is not all properties are 3d so we can't build a single numpy array from that. Also the x,y,z arrays are already read/generated directly from `lammps.read(...)` and `tools.readxyz(...)` as separate arrays.

BTW, I've removed these lines from the example since np.stack already treats them as arrays.

XzzX (approved):

PR #330, opened 2021-04-15
option to automatically look for and use ccache

XzzX:
Are you interested in automatic ccache detection?  
Implementation according to:
https://cliutils.gitlab.io/modern-cmake/chapters/features/utilities.html


junghans (commented):
Hmm, I am not sure why this is needed.

```
export PATH=/usr/lib64/ccache:${PATH}
```
is the usual way to enable ccache.

On the other hand it won't hurt.

XzzX:
> Hmm, I am not sure why this is needed.
> 
> ```
> export PATH=/usr/lib64/ccache:${PATH}
> ```
> 
> is the usual way to enable ccache.
> 
> On the other hand it won't hurt.

The point of this PR is not to make finding `ccache` easier. In a default installation it will be in the search path anyway. It makes enabling `ccache` easier, i.e. removing the need for `export CXX="ccache g++"` or something similar.

You can use `cmake -DUSE_CCACHE=ON` and cmake will locate and configure it correctly if it is installed.

I think it is personal preference but might come in handy for example in IDEs if accessing environment variables is tricky.

junghans:
Sure, as I said it won't hurt.

junghans (approved):

junghans:
We should enable ccache support in github actions as well.
```
    env:
      CCACHE_DIR: "${{ github.workspace }}/.ccache"
      CCACHE_MAXSIZE: "10G"
      - name: Cache ccache
        uses: actions/cache@v2
        with:
          path: ${{ env.CCACHE_DIR }}
          key: ccache-${{ matrix.distro }}-${{github.run_id}}
          restore-keys: ccache-${{ matrix.distro }}
```

XzzX:
CCache support in GitHub actions can be enabled separately.

PR #332, opened 2021-04-15
replace travis ci with github actions

### LINKED ISSUE #327: CI is failing for current master
fedora images do not have a proper mpi4py installation:  
 - [ ] https://travis-ci.org/github/espressopp/espressopp/jobs/766333270#L346

opensuse images have multiple issues:
  
- [ ] h5py build without mpi support
https://travis-ci.org/github/espressopp/espressopp/jobs/766333272#L5210
- [ ] problems while writing output files
https://travis-ci.org/github/espressopp/espressopp/jobs/766333272#L5160

XzzX:
- [x] introduce github actions ci
- [x] update README badge
- [x] remove travis ci

closes #327


.github/workflows/cmake.yml
```diff
@@ -0,0 +1,53 @@
...
+    - name: Build
+      working-directory: ${{github.workspace}}/build
+      shell: bash
+      # Execute the build.  You can specify a specific target with "--target <NAME>"
+      run: cmake --build . --target all -j 2
+
+    - name: Test
+      working-directory: ${{github.workspace}}/build
+      env:
+          PYTHONPATH: ${{github.workspace}}/build:${{github.workspace}}/build/contrib
```
junghans, line 49: That should not be needed as CMake sets these paths automatically, see https://github.com/espressopp/espressopp/blob/05ae583ddabb47e959f867fc0b7539c6af2bd599/CMakeLists.txt#L176

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,53 @@
...
+  build:
+    strategy:
+      matrix:
+        BUILD_TYPE: [Release, Debug]
+      
+    # The CMake configure and build commands are platform agnostic and should work equally
+    # well on Windows or Mac.  You can convert this to a matrix build if you need
+    # cross-platform coverage.
+    # See: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
+    runs-on: ubuntu-latest
```
junghans, line 16: Later we should use our containers here:
```
 distro: ['opensuse', 'ubuntu_rolling', 'fedora', 'ubuntu', 'ubuntu_mpich', 'fedora_mpich']
container: espresso/buildenv:${{ matrix.distro }}
```

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,53 @@
+name: CMake
```
junghans, line 1:
```suggestion
name: CI
```

.travis.yml
```diff
@@ -1,86 +0,0 @@
...
-        - stage: coverage
-          env: DISTRO=ubuntu COVERAGE=ON
-        - stage: deploy
-          if: type != pull_request
-          env: DISTRO=ubuntu BUILD_UG=ON XTC=ON
-
-after_success:
-  - if [[ "${TRAVIS_BUILD_STAGE_NAME}" != [dD]eploy ]]; then travis_terminate 0; fi
-  - shopt -s extglob && [[ ${TRAVIS_BRANCH} = @(master|v[12].*) ]] && DEPLOY=yes
-  - if [[ ${DOCKER_USERNAME} && ${DOCKER_PASSWORD} && ${TRAVIS_PULL_REQUEST} == false && ${DEPLOY} ]]; then
```
junghans, line 42: We need to replace the docker deploy as well.

.github/workflows/cmake.yml
XzzX: Ah, ok. But this is not compatible with `EXTERNAL_MPI4PY=OFF` as the `contrib` directory is missing. We could add the `contrib` directory here as well. However, I vote for removing the `TEST_ENV` entirely, because running the examples does not have the pre set-up environment. This way running tests and examples would need the same preparation steps.

.travis.yml
XzzX: Is the docker deploy run for EVERY master commit or only for releases?

.travis.yml
```diff
@@ -1,86 +0,0 @@
...
-      fi;
-      rm -rvf doc;
-      cd html;
-      if [[ ${DEPLOY} && ${encrypted_194b3d1e9306_key} && ${encrypted_194b3d1e9306_iv} && ${TRAVIS_PULL_REQUEST} == false ]]; then
-        git config --global user.name "Automatic Deployment (Travis CI)";
-        git config --global user.email "espp-devel@listserv.mpip-mainz.mpg.de";
-        git add --all && git commit -m "Documentation Update";
-        openssl aes-256-cbc -K $encrypted_194b3d1e9306_key -iv $encrypted_194b3d1e9306_iv -in $HOME/docker/espressopp/doc/ug/deploy.enc -out ~/.ssh/id_rsa -d;
-        chmod 600 ~/.ssh/id_rsa;
-        git push git@github.com:espressopp/espressopp.github.io.git master;
```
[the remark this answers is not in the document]
XzzX: Same question as for the docker container update. Do you want to update it with every master commit?

.github/workflows/cmake.yml
XzzX: If we want to maintain the containers, yes. Is there much difference between the different OSes? Or do you want to test if `espressopp` can be compiled with the packaged libraries?

.github/workflows/cmake.yml
XzzX: This is the default name of the provided template :) I will rename it and also the file.

.travis.yml
junghans: yes

.travis.yml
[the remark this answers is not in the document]
junghans: Every commit to master.

.github/workflows/cmake.yml
junghans: All distro have different version of all kind of packages and compilers hence it is good to test all of them....

.github/workflows/cmake.yml
junghans: Huh, we had that in v2 https://github.com/espressopp/espressopp/blob/7a8e90426a53ecaf99225e682563542b6b1701cd/CMakeLists.txt#L190-L194

junghans: Not sure why it got removed.

junghans:
Honestly, I would put the docker build in a separate file.

.github/workflows/cmake.yml
XzzX: done

.github/workflows/cmake.yml
XzzX: What about removing it completely for consistency reasons?

XzzX:
> Honestly, I would put the docker build in a separate file.

I am just testing. I need someone explaining me your docker file. I expected something "simpler".

junghans:
> > Honestly, I would put the docker build in a separate file.
> 
> I am just testing. I need someone explaining me your docker file. I expected something "simpler".

Yes, that was due to Travis, now we could shrink to it just the version we want to deploy to docker hub.

XzzX:
Ok, I will create a version how I expected it.

XzzX:
@junghans Please have a look at https://github.com/XzzX/espressopp/blob/github-ci/docker/Dockerfile

ToDo:
- [ ] restrict to master branch pushes
- [ ] provide tag and credentials for DockerHub

.github/workflows/cmake.yml
XzzX: I am afraid custom Docker images are only available either by using self-hosted runners or running docker inside the VM. https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on
I do not really like both solutions.

junghans (dismissed):
I would say let's merge this and make a todo issue for the remaining things

- [ ] Doc deploy
- [ ] Docker deploy
- [ ] different distro / compilers

junghans:
> @junghans Please have a look at https://github.com/XzzX/espressopp/blob/github-ci/docker/Dockerfile

I would use modern cmake and drop `cd`, `mkdir`:
```
cmake -B builddir -S espressopp
cmake --build builddir
cmake --build builddir --target test 
cmake --install builddir
```

> 
> ToDo:
> 
> * [ ]  restrict to master branch pushes

I will unrestricted this shortly and then we can go from there and re-enable once merged.
Will give you permissions too.

> * [ ]  provide tag and credentials for DockerHub

Docker push v2 has an issue that prevent big layer to be pushed (https://github.com/docker/build-push-action/issues/251) so I guess we will have to stick to v1 for now.

junghans (approved):

docker/Dockerfile
```diff
@@ -1,47 +1,9 @@
...
-    $HOME/.local/bin/codecov --gcov-exec "llvm-cov gcov" > /dev/null; \
-  else \
-    $HOME/.local/bin/codecov > /dev/null; \
-  fi; \
-fi && cd -
-USER root
-RUN make install
-USER espressopp
-WORKDIR ../examples
+RUN git clone https://github.com/espressopp/espressopp.git --depth 1
```
junghans, line 3: This is wrong, as it would always build master instead of current commit.

Use the checkout action in docker workflow.

docker/Dockerfile
```diff
@@ -1,47 +1,9 @@
...
+RUN mkdir espressopp-build \
+ && cd espressopp-build \
+ && pwd && ls -lah \
+ && cmake ../espressopp -DCMAKE_BUILD_TYPE=Release -DEXTERNAL_MPI4PY=OFF \
+ && make -j2
```
junghans, line 8:
```suggestion
RUN cmake -B espressopp-build -S ../espressopp -DCMAKE_BUILD_TYPE=Release -DEXTERNAL_MPI4PY=OFF \
 && make -C espressopp-build -j2
```

.github/workflows/cmake.yml
junghans: I would really like the test to work with having to set the `PYTHONPATH` externally, so we should add that back.

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,35 @@
...
+
+jobs:
+
+  documentation:
+    runs-on: espressopp/buildenv:ubuntu
+
+    steps:
+    - name: build documentation
+      run: |
+        git clone https://github.com/espressopp/espressopp.git
```
junghans, line 17: Use the checkout action instead.

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,35 @@
...
+        mkdir espressopp-build
+        cd espressopp-build
+        cmake ../espressopp
+        make doc
+
+    - name: update docu repo
+      run: |
+        git clone --depth 1 https://github.com/espressopp/espressopp.github.io.git html;
```
junghans, line 25: use the checkout action instead.

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,35 @@
...
+        mv espressopp-build/doc/ug/_build/latex/ESPResSo++.pdf html/latest/;
+        mv espressopp-build/doc/ug/_build/html/* html/latest/;
+        git config --global user.name "Automatic Deployment (Travis CI)";
+        git config --global user.email "espp-devel@listserv.mpip-mainz.mpg.de";
+        git add --all && git commit -m "Documentation Update";
+
+    - name: deploy documentation
+      run: |
+        cd html
+        openssl aes-256-cbc -K $encrypted_194b3d1e9306_key -iv $encrypted_194b3d1e9306_iv -in $HOME/docker/espressopp/doc/ug/deploy.enc -out ~/.ssh/id_rsa -d;
```
junghans, line 38: In Github actions we either should add a PAT in a secret to do the deploy or generate a new deploy ssh key and store it private key in a secret.
```
echo ${{ SECRET.private_deploy_key }} | base64 -d > ~/.ssh/id_rsa
```

.github/workflows/cmake.yml
XzzX: I would prefer the other way, but well, not a big issue. I will add it.

docker/Dockerfile
XzzX: This is the docker file, not the github actions file. I can link the sources from github actions into the docker container, yes, but why. Do you want to have a deployed docker container for every work in progress branch?!? I thought docker containers are meant for end users who do not want to build espressopp themselves? Furthermore, this will produce errors in forked repositories which do not have the secrets. I think I need some clarification what your containers are meant for.

docker/Dockerfile
XzzX: done

.github/workflows/documentation.yml
XzzX: Same as for docker containers. Do you want to publish a documentation for every WIP branch?

docker/Dockerfile
junghans: Well we want to test if a certain set of changes still builds, right? Otherwise we might merge a PR just to realize the docker build is not broken. (Same for the documentation below). One can set `push` to false for no push events.

https://github.com/votca/votca/blob/master/.github/workflows/docker-build.yml#L34

.github/workflows/documentation.yml
junghans: We want to test if the documentation still builds before it gets merged.

docker/Dockerfile
XzzX: Building, yes. But do you also want to deploy? In particular, this will make all builds fail on forked repositories where the secrets are not available. I think deploy should be restricted to the master branch.

junghans: Yes, I would only deploy it for the main repo and pushed to master (and maybe nightly builds)

.github/workflows/cmake.yml
XzzX: Different compilers and OSes will be added in a separate MR.

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,52 @@
...
+    - name: build documentation
+      run: |
+        cmake -S espressopp -B espressopp-build
+        cmake --build espressopp-build --target doc
+
+    - uses: actions/checkout@v2
+      with:
+        repository: 'espressopp/espressopp.github.io.git'
+        path: 'docu'
+      if: github.ref == 'refs/heads/master'
```
junghans, line 28: Isn't that also true if the target of a PR is master?

Maybe:
```
if: github.ref == 'refs/heads/master' && ( github.event_name == 'push' ||  github.event_name == 'schedule' )
```

.github/workflows/documentation.yml
junghans: https://github.community/t/github-actions-ci-how-to-use-store-deploy-key-to-download-from-another-private-repo/16113/7

.github/workflows/documentation.yml
junghans: And honestly I would always do everything but the push.

.github/workflows/Dockerfile
```diff
@@ -0,0 +1,6 @@
+from espressopp/buildenv:ubuntu
+
+COPY espressopp/ /home/espressopp/espressopp
+RUN cmake -S espressopp -B espressopp-build -DCMAKE_BUILD_TYPE=Release -DEXTERNAL_MPI4PY=OFF
```
junghans, line 4: Does the MPI4PY in the ubuntu image not work?

.github/workflows/Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+RUN cmake --build espressopp-build -j2
```
junghans, line 5: and shouldn't we install it too?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,52 @@
...
+    - uses: actions/checkout@v2
+    
+    # install dependencies
+    - name: dependencies
+      run: sudo apt-get update && sudo apt-get install -yq openmpi-bin libfftw3-dev python3-dev libboost-all-dev python3-h5py
+
+    - name: Create Build Environment
+      # Some projects don't allow in-source building, so create a separate build directory
+      # We'll use this as our working directory for all subsequent commands
+      run: cmake -E make_directory ${{github.workspace}}/build
```
junghans, line 28: You could drop that step if you run `cmake -B ..` below.

junghans (commented):
Only minor nitpicks left.

.github/workflows/documentation.yml
XzzX: done

.github/workflows/documentation.yml
XzzX: It triggers for all master branch pushes (main repo and forks). Does not trigger for PRs. Already tested.  
https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

.github/workflows/Dockerfile
XzzX: I do not even know why you bundle mpi4py. @stuehn mentioned some hacky conversion between mpi4py communicator and c++ communicator. Perhaps a fixed version is needed for this conversion to work reliably.

.github/workflows/Dockerfile
XzzX: I thought linking to the build folder is better as this also allows some development/rebuild in the docker container. If we use an installed version you need to reinstall.

.github/workflows/ci.yml
XzzX: Yes, still old template. I will rework it.

.github/workflows/Dockerfile
XzzX: I added an install step to the CI runs.

XzzX: Install is currently broken. Will be a separate issue.

.github/workflows/documentation.yml
XzzX: done

XzzX:
@junghans If you want, you can have another look at the PR. From my side the PR is now finished.

I recombined all actions into one file as I see it as a validation action for a commit. Since all actions were triggered under the same conditions I think this makes more sense and also simplifies for example cancelling.

I also reworked the documentation step. It is now build in the build directory and not in the source directory. I think this is more natural.

Currently I cannot test the docker deploy as I cannot extract the secrets from the main repo. We will se if it works once the PR is merged.

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,112 @@
...
+      - name: Test
+        working-directory: ${{github.workspace}}/build
+        run: ctest --rerun-failed --output-on-failure -j 2
+
+  documentation:
+    runs-on: ubuntu-latest
+    environment: CI
+    steps:
+      - name: install dependencies
+        # very ugly but did not get line breaks working...
```
junghans, line 40: I think you can do `run: |` here and then use `\` at the EOL below.

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,112 @@
...
+          git config --global user.name "Automatic Deployment (Travis CI)";
+          git config --global user.email "espp-devel@listserv.mpip-mainz.mpg.de";
+          git add --all && git commit -m "Documentation Update";
+
+      - name: register deploy key
+        env:
+          SSH_AUTH_SOCK: /tmp/ssh_agent.sock
+        run: |
+          ssh-agent -a $SSH_AUTH_SOCK > /dev/null
+          ssh-add - <<< "${{ secrets.PAGES_KEY }}"
```
junghans, line 77: Is there any advantage of `ssh-agent` over just put it in a file?

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,112 @@
+name: VALIDATE
+
+on:
+  push:
```
junghans, line 4: Do we really want to trigger this for every brunch?

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,112 @@
+name: VALIDATE
```
junghans, line 1: Will the job run in parallel or sequentially when they are all in one file?

doc/sphinx/index.rst
```diff
@@ -15,8 +15,6 @@ Welcome to the user documentation of ESPResSo++
    tutorial.rst
    classdescription.rst
 
-   doxygen 
-
```
junghans, line 19: Are we not integrating that anymore?

junghans (approved):
Wondering why you moved the docs around, but ok.

doc/sphinx/index.rst
XzzX: I removed it, since doxygen comments are rarely used in the code base and and I do not see much use in a plain listing of functions without explanations. Any decent IDE can give you the same. Doxygen is still built, so if you need it I can reenable it.

.github/workflows/validate.yaml
XzzX: Everything runs in parallel.

.github/workflows/validate.yaml
XzzX: I would say so. I also want to have the checks during development. Fixing errors continuously is easier than fixing a large chunk at the end.

.github/workflows/validate.yaml
XzzX: It got it working faster than the other way ;) I do not think there is much difference.

.github/workflows/validate.yaml
XzzX: I tried with `\` but not with `|`. I will check that.

XzzX: perfect, working

.github/workflows/validate.yaml
junghans: ok, I would just open an PR then, but ok, we can always disable if it becomes too much.

doc/sphinx/index.rst
junghans: ok

PR #334, opened 2021-04-22
Extend vectorization

jnvance:
- Extends the use of vectorizable data layout (SOA) to entire integration scheme
- Submodule `espressopp.vec` added to replace `vectorization`
- Offload particle data from array-of-structures `ParticleList` to structure-of-arrays `ParticleArray` only during initialization and resort
- Performs Lennard-Jones and Polymer Melt simulations
- Supports analysis tools before and after integrator.run

TODO:
- [x] Implement general tests
- [x] Implement tests covering all classes 
- [x] Merge changes to CI
- [x] Remove AOS implementation (packed4)
- [x] Create ParticleArray iterator
- [x] Remove espressopp.vectorization


src/vec/include/packed4.hpp
```diff
@@ -0,0 +1,86 @@
...
+  struct Real3DInt {
+    real x;
+    real y;
+    real z;
+    lint t;
+
+    Real3DInt()
+    {
+      // for padding to work, real must be 64 bit floating point type
+      static_assert(sizeof(real)==sizeof(lint),
```
XzzX, line 37: Seems odd to check the size of a floating point type against the size of an integer type. Why not `sizeof(double)`.

This check will de facto eliminate the possibility to build in single precision rendering the choice irrelevant.

jnvance: Will be removing the AOS implementation so this part won't be needed.

src/vec/FixedPairList.cpp
```diff
@@ -0,0 +1,345 @@
...
+    return ret;
+  }
+
+  python::list FixedPairList::getAllBonds()
+  {
+    std::vector<size_t> local_bonds;
+    std::vector<std::vector<size_t> > global_bonds;
+    python::list bonds;
+
+    for (GlobalPairs::const_iterator it = globalPairs.begin(); it != globalPairs.end(); it++) {
```
XzzX, line 172: I prefer `auto` for iterators. This also applies for the other iterator-based loops.
```suggestion
    for (auto it = globalPairs.cbegin(); it != globalPairs.cend(); it++) {
```

src/vec/FixedPairList.cpp
```diff
@@ -0,0 +1,345 @@
...
+      mpi::gather(*system.comm, local_bonds, global_bonds, 0);
+      python::tuple bond;
+
+      for (std::vector<std::vector<size_t> >::iterator it = global_bonds.begin();
+           it != global_bonds.end(); ++it) {
+        for (std::vector<size_t>::iterator iit = it->begin(); iit != it->end();) {
+          size_t pid1 = *(iit++);
+          size_t pid2 = *(iit++);
+          bonds.append(python::make_tuple(pid1, pid2));
+        }
```
XzzX, line 187: I think readability can be improved by
```suggestion
        for (auto iit = it->begin(); iit != it->end(); iit += 2) {
          size_t pid1 = *(iit);
          size_t pid2 = *(iit + 1);
          bonds.append(python::make_tuple(pid1, pid2));
        }
```

src/vec/FixedPairList.cpp
```diff
@@ -0,0 +1,345 @@
...
+    python::list bonds;
+
+    for (GlobalPairs::const_iterator it = globalPairs.begin(); it != globalPairs.end(); it++) {
+      local_bonds.push_back(it->first);
+      local_bonds.push_back(it->second);
+    }
+    System& system = vectorization->getSystemRef();
+    if (system.comm->rank() == 0) {
+      mpi::gather(*system.comm, local_bonds, global_bonds, 0);
+      python::tuple bond;
```
XzzX, line 179: `bond` unused?

src/vec/FixedPairList.cpp
```diff
@@ -0,0 +1,345 @@
...
+      LOG4ESPP_DEBUG(theLogger, "send particle with pid " << pid << ", find pairs");
+
+      // find all pairs that involve this particle
+
+      int n = globalPairs.count(pid);
+
+      if (n > 0) {
+        std::pair<GlobalPairs::const_iterator,
+          GlobalPairs::const_iterator> equalRange
+          = globalPairs.equal_range(pid);
```
XzzX, line 211:
```suggestion
        auto equalRange = globalPairs.equal_range(pid);
```

src/vec/FixedPairList.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+  std::vector<size_t> FixedPairList::getPairList()
+  {
+    std::vector<size_t> ret;
+    for (GlobalPairs::const_iterator it = globalPairs.begin(); it != globalPairs.end(); it++) {
+      ret.push_back(it->first);
+      ret.push_back(it->second);
+    }
+    return ret;
+  }
```
XzzX, line 164: Should be a free function converting general pairs into a linearized version.

src/vec/FixedPairList.cpp
```diff
@@ -0,0 +1,345 @@
...
+      if (equalRange.first == globalPairs.end()) {
+        // if it hasn't, insert the new pair
+        globalPairs.insert(std::make_pair(pid1, pid2));
+      }
+      else {
+        // otherwise test whether the pair already exists
+        for (GlobalPairs::const_iterator it = equalRange.first; it != equalRange.second; ++it) {
+          if (it->second == pid2) {
+            // Pair already exists, generate error!
+            throw std::runtime_error("Duplicate pair inserted");
```
XzzX, line 131: Have you included this check also in the other versions of FixedPairList?

Comment is unnecessary at this point.

src/vec/FixedPairList.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+    if(returnVal){
+      // add the pair locally
+      this->push_back({p1,p2});
+
+      // ADD THE GLOBAL PAIR
+      // see whether the particle already has pairs
+      std::pair<GlobalPairs::const_iterator,
+        GlobalPairs::const_iterator> equalRange
+        = globalPairs.equal_range(pid1);
```
XzzX, line 121:
```suggestion
      auto equalRange = globalPairs.equal_range(pid1);
```

src/vec/FixedPairList.cpp
```diff
@@ -0,0 +1,345 @@
...
+    if (pid1 > pid2)
+      std::swap(pid1, pid2);
+
+    auto const& storageVec = vectorization->storageVec;
+    size_t const p1 = storageVec->lookupRealParticleVec(pid1);
+    size_t const p2 = storageVec->lookupLocalParticleVec(pid2);
+
+    if(p1==VEC_PARTICLE_NOT_FOUND) {
+      // Particle does not exist here, return false
+      returnVal=false;
```
XzzX, line 106:
```suggestion
      return false;
```

Makes following `if` unnecessary.

XzzX (commented):
How detailed do you want the comments to be?

src/vec/FixedPairList.cpp
jnvance: This is needed to trigger LOG4ESPP_DEBUG before return as implemented in original version.

jnvance:
> How detailed do you want the comments to be?

Thanks for the comments. Most of these refer to the code parts that I just copied from the original implementation in src/FixedPairList.cpp. Anyway, I'll implement your suggestion of using auto on other source files.

src/vec/FixedPairList.cpp
jnvance: Kept it as it is because I'm not sure the same functionality is used anywhere else aside from local_bonds in getAllBonds(). Unless we generalize it to triples, quadruples, etc.

src/vec/FixedPairList.cpp
jnvance: > Have you included this check also in the other versions of FixedPairList?
No, still a TODO there.

> Comment is unnecessary at this point.
Edited.

src/vec/FixedPairList.cpp
XzzX: Why not move the `LOG4ESPP_DEBUG` to this place as well?

XzzX:
> > How detailed do you want the comments to be?
> 
> Thanks for the comments. Most of these refer to the code parts that I just copied from the original implementation in src/FixedPairList.cpp. Anyway, I'll implement your suggestion of using auto on other source files.

This is why I do not like the idea of duplicating source code. This is known as the `DRY` principle, `do not repeat yourself`. If we want to have three copies of `FixedPairList` (not to forget the Adress version) we have to maintain them all, and fix bugs in all of them.

src/vec/FixedPairList.cpp
```diff
@@ -0,0 +1,336 @@
...
+    // send the list
+    buf.write(toSend);
+    LOG4ESPP_INFO(theLogger, "prepared fixed pair list before send particles");
+  }
+
+  void FixedPairList::afterRecvParticles(ParticleList &pl, InBuffer& buf)
+  {
+    std::vector< size_t > received;
+    int n;
+    size_t pid1, pid2;
```
XzzX, line 231: Do not use uninitialized variables! Possible source of undefined behavior. Declare them where you need them.

src/vec/FixedPairList.cpp
```diff
@@ -0,0 +1,336 @@
...
+    auto it = globalPairs.begin();
+    // receive the bond list
+    buf.read(received);
+    int size = received.size(); int i = 0;
+    while (i < size) {
+      // unpack the list
+      pid1 = received[i++];
```
XzzX, line 238: Matter of taste, but why using pre/post-increment in expresseions. In my opinion this is a unnecessary source of confusion and it is error prone.

jnvance:
> > > How detailed do you want the comments to be?
> > 
> > 
> > Thanks for the comments. Most of these refer to the code parts that I just copied from the original implementation in src/FixedPairList.cpp. Anyway, I'll implement your suggestion of using auto on other source files.
> 
> This is why I do not like the idea of duplicating source code. This is known as the `DRY` principle, `do not repeat yourself`. If we want to have three copies of `FixedPairList` (not to forget the Adress version) we have to maintain them all, and fix bugs in all of them.

What I meant was other source files in `vec` ([4cfb90e](https://github.com/espressopp/espressopp/pull/334/commits/4cfb90e8515b0909d190013ffb5ac5b6ba206afb)). Since we are most likely moving to the SOA, there's probably no need to implement the style changes in the original source files.

src/vec/FixedPairList.cpp
jnvance: Right, and I've also removed the returnVal variable entirely

junghans:
I like it, however I am still wondering if this is a little too home-grown if we should just use an existing framework like [kokkos](https://github.com/kokkos/kokkos), [Cabana](https://github.com/ECP-copa/Cabana) or [ispc](https://github.com/ispc/ispc) instead.

jnvance:
> I like it, however I am still wondering if this is a little too home-grown if we should just use an existing framework like [kokkos](https://github.com/kokkos/kokkos), [Cabana](https://github.com/ECP-copa/Cabana) or [ispc](https://github.com/ispc/ispc) instead.

This is just a starting point to see how much can already be gained just from changing the Particle data layout to SOA and applying some simple vectorization pragmas (e.g. overall 3.2x for LJ and 2x for polymer melt on single Skylake node). We are planning to look into Kokkos/Cabana especially for the GPU integration.

XzzX:
Do you also plan to implement unit tests for each vectorized loop? I think an integration test like lennard_jones or poly_melt cannot really catch everything.

jnvance:
> Do you also plan to implement unit tests for each vectorized loop? I think an integration test like lennard_jones or poly_melt cannot really catch everything.

That's probably a good idea. I'll probably implement one for the integrator and interactions.

src/vec/FixedPairList.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+    for (auto it = globalPairs.cbegin(); it != globalPairs.cend(); it++) {
+      local_bonds.push_back(it->first);
+      local_bonds.push_back(it->second);
+    }
+    System& system = vectorization->getSystemRef();
+    if (system.comm->rank() == 0) {
+      mpi::gather(*system.comm, local_bonds, global_bonds, 0);
+      for (auto it = global_bonds.begin(); it != global_bonds.end(); ++it) {
+        for (auto iit = it->begin(); iit != it->end(); iit += 2) {
```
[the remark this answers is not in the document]
jnvance: Including src/checks.hpp causes an error: 

/__w/espressopp/espressopp/src/vec/FixedPairList.cpp(356): error #2196: routine is both "inline" and "noinline"

src/vec/FixedPairList.cpp
jnvance: The use of `received[i++]` in this usage makes sense so I left it as is.

src/vec/FixedPairList.cpp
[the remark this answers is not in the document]
XzzX: I noticed this today morning. No idea why. Only happens with `icpc`.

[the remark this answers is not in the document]
XzzX: The internet is also pretty silent about this Intel warning...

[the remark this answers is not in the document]
junghans: https://community.intel.com/t5/Intel-C-Compiler/Inline-and-no-inline-warning/td-p/1216764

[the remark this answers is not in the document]
XzzX: This thread has no solution apart from deactivating it or using `icpx`.

I have exactly one function marked `inline` which I could remove. The warning, however, persists.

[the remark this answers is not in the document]
jnvance: The warning seems to come from `#include <boost/stacktrace.hpp>`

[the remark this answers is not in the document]
XzzX: Good find. Now we need a solution. Removing the stacktrace?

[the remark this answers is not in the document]
jnvance: Yeah, I think we only really need the source code file and line number to figure out most assertion errors.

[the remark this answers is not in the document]
XzzX: If the error is in a very common file a stacktrace can be really helpful.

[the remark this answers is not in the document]
jnvance: If we want to keep it and have no warnings, one option would be:
```cpp
#pragma warning(push)
#pragma warning(disable:2196)
#include <boost/stacktrace.hpp>
```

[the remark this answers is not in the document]
XzzX: I am fine with both.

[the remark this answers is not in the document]
XzzX: Let's remove it for now: #397

PR #336, opened 2021-04-23
Delete .gitlab-ci.yml

junghans:


XzzX:
Any reason why we support Gitlab at all? I suggest removing the Gitlab repository completely and to bundle all activity on Github. In fact, we do not really want issues or pull requests on Gitlab, do we?

XzzX (approved):

PR #337, opened 2021-04-23
cleanup & extension of gitignore

XzzX:
I vote for strictly separating source and build directories.
This would make some ignores obsolete, however, requires a change
in philosophy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,27 +1,49 @@
-*~
-Makefile
-*.o
-*.so
-*.pyc
-COPYING
-INSTALL
+*build*
+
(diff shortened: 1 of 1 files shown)


junghans (changes requested):
Can we leave `*~` in there?

XzzX:
Yes. What is the purpose of this entry?

junghans:
> Yes. What is the purpose of this entry?

Vim backup files

XzzX:
done.

junghans (approved):

PR #344, opened 2021-04-23
CI: test more distros

### LINKED ISSUE #343: enable CCache in GitHub actions

junghans:
To Do:

- [x] add intel and intel-oneapi builds

Fix #343

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -11,58 +11,54 @@ jobs:
     strategy:
       matrix:
         BUILD_TYPE: [Release, Debug]
+        DISTRO: ['fedora']
+        CXX: ['g++', 'clang++']
+        EXTERNAL_MPI4PY: ['On']
+        include:
+          - BUILD_TYPE: 'Debug'
+            DISTRO: 'fedora_intel'
+            CXX: 'icpc'
+            EXTERNAL_MPI4PY: 'On'
+          - BUILD_TYPE: 'Debug'
+            DISTRO: 'fedora_intel-oneapi'
+            CXX: 'icpx'
+            EXTERNAL_MPI4PY: 'On'
 
     runs-on: ubuntu-latest
+    container: espressopp/buildenv:${{ matrix.distro }}
     steps:
+      - uses: actions/cache@v2
+        with:
+          path: ~/.ccache
+          key: ${{ matrix.DISTRO }}-${{ matrix.CXX }}-${{ matrix.BUILD_TYPE }}-${{ github.run_id }}
+          restore-keys: ${{ matrix.DISTRO }}-${{ matrix.CXX }}-${{ matrix.BUILD_TYPE }}
+
       - uses: actions/checkout@v2
 
-      - name: Dependencies
+      - name: Get trail license
+        if: ${{ matrix.CXX == 'icpc' }}
         run: |
-          sudo apt-get update
-          sudo apt-get install -yq \
-          libboost-all-dev \
-          libfftw3-dev \
-          openmpi-bin \
-          python3-dev \
-          python3-h5py
-
-
-
+          mkdir ~/Licenses
+          curl https://dynamicinstaller.intel.com/api/v2/license > ~/Licenses/intel.lic
+      
       - name: Configure
-        run: cmake -S ${{github.workspace}} -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{ matrix.BUILD_TYPE }} -DEXTERNAL_MPI4PY=OFF
+        run: cmake -S . -B build -DCMAKE_BUILD_TYPE=${{ matrix.BUILD_TYPE }} -DCMAKE_CXX_COMPILER=${{ matrix.CXX }} -DEXTERNAL_MPI4PY=${{ matrix.EXTERNAL_MPI4PY }}
 
       - name: Build
-        run: cmake --build ${{github.workspace}}/build --target all -j 2
+        run: cmake --build build --verbose --target all -j 2
 
       #    - name: Install
-      #      run: cmake --install ${{github.workspace}}/build
+      #      run: cmake --install build
 
       - name: Test
-        working-directory: ${{github.workspace}}/build
+        working-directory: build
         run: ctest --rerun-failed --output-on-failure -j 2
 
   documentation:
     runs-on: ubuntu-latest
     environment: CI
+    container: espressopp/buildenv:fedora
     steps:
-      - name: install dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get install -yq \
-          doxygen \
-          ghostscript \
-          graphviz \
-          latexmk \
-          libboost-all-dev \
-          libfftw3-dev \
-          python3-dev \
-          python3-h5py \
-          openmpi-bin \
-          sphinx-common \
-          texlive-latex-base \
-          texlive-latex-extra \
-          texlive-latex-recommended
-
       - uses: actions/checkout@v2
         with:
           path: 'espressopp'
@@ -88,7 +84,7 @@ jobs:
       - name: commit
         working-directory: ${{github.workspace}}/html
         run: |
-          git config --global user.name "Automatic Deployment (Travis CI)";
+          git config --global user.name "Automatic Deployment (GitHub Action)";
           git config --global user.email "espp-devel@listserv.mpip-mainz.mpg.de";
           git add --all && git commit -m "Documentation Update";
 
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -1,0 +1,53 @@
+name: WEEKLY
+
+on:
+  schedule:
+    - cron:  '0 0 * * SAT'
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        BUILD_TYPE: [Release, Debug]
+        DISTRO: ['fedora', 'fedora_rawhide', 'fedora_mpich', 'opensuse', 'ubuntu', 'ubuntu_devel', 'ubuntu_rolling', 'ubuntu_mpich']
+        CXX: ['g++', 'clang++']
+        EXTERNAL_MPI4PY: ['On','Off']
+        include:
+          - BUILD_TYPE: 'Debug'
+            DISTRO: 'fedora_intel'
+            CXX: 'icpc'
+            EXTERNAL_MPI4PY: 'On'
+          - BUILD_TYPE: 'Debug'
+            DISTRO: 'fedora_intel-oneapi'
+            CXX: 'icpx'
+            EXTERNAL_MPI4PY: 'On'
+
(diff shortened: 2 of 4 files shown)


XzzX:
I am unsure if we should run all builds on every commit / PR. Perhaps we can move some checks to weekly.

junghans:
> I am unsure if we should run all builds on every commit / PR. Perhaps we can move some checks to weekly.

Yeah, 8 x 2 x 2 might be too much, but let's see what breaks and then down-select.

.github/workflows/validate.yml
```diff
@@ -11,58 +11,54 @@ jobs:
     strategy:
       matrix:
         BUILD_TYPE: [Release, Debug]
+        DISTRO: ['fedora', 'fedora_rawhide', 'fedora_mpich', 'opensuse', 'ubuntu', 'ubuntu_devel', 'ubuntu_rolling', 'ubuntu_mpich']
+        CXX: ['g++', 'clang++']
+        EXTERNAL_MPI4PY: ['On','Off']
```
XzzX, line 16: Goal should be to get rid of this switch entirely. Either we should stick to a preinstalled one if we do not have special requirements or we should always use ours, if we have special requirements. I do not see a point in having it switchable.

.github/workflows/validate.yml
```diff
@@ -11,58 +11,54 @@ jobs:
...
+        include:
+          - BUILD_TYPE: 'Debug'
+            DISTRO: 'fedora_intel'
+            CXX: 'icpc'
+            EXTERNAL_MPI4PY: 'On'
```
XzzX, line 21: Why is Intel only available in fedora?

junghans: Because I only installed it in the Fedora container.

I am sure it is possible under ubuntu as well.

.github/workflows/validate.yml
junghans: It seems to currently work with all distribution packages, so we should just drop our bundled version.

.github/workflows/validate.yml
```diff
@@ -11,58 +11,54 @@ jobs:
...
 
       - name: Test
-        working-directory: ${{github.workspace}}/build
+        working-directory: build
         run: ctest --rerun-failed --output-on-failure -j 2
 
   documentation:
     runs-on: ubuntu-latest
     environment: CI
+    container: espressopp/buildenv:fedora
```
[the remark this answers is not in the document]
junghans: The GitHub vm has a wild mix of ubuntu packages and custom craft installed in /opt. Hence it is a bit tricky to reproduce any errors locally, hence I personally prefer to use containers whereever possible. That makes the debugging much easier as one can just get the whole container stack from docker hub.

junghans:
> After thinking about it I propose the following structure:
> 
> on commit (remove on PR as this is already covered by on commit):
> 
> * based on ubuntu-latest github actions vm
> * g++, clang, Intel
> * debug, release
> * documentation
> 
> weekly, deployment:
> 
> * build&push docker containers for all releases
> * build&push documentation
> 
> weekly, compatibility:
> 
> * builds for various distros
> * different MPI versions
> 
> I think this is a good compromise between resource consumption and coverage.
> @junghans Your thoughts?

Like I said above I don’t like the GitHub action vm, other than that I am ok with the plan. I would maybe push the docker image on every merge to master to avoid confusion.

However that might mean we will have to split the workflow file as we cannot keep different schedules in one file.

.github/workflows/validate.yml
XzzX: I `mpi4py` built against `mpi`? If so, can we ensure that this is always correct? Also on clusters with multiple `mpi` versions?

junghans: Well I think that is the same with all of our other dependencies, e.g. hdf5, boost-mpi.

junghans: On a cluster I would use spack install espressopp, but the package needs an update too.

junghans:
And the intel compiler package  takes long to download and install, so we want to use the container for that.

.github/workflows/validate.yml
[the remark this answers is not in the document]
XzzX: Oh, I did not know that. Very valid point. We should run everything in containers then.

.github/workflows/validate.yml
XzzX: Hmm, for hdf5 and boost you typically can load special packages for each `mpi` version. I have not seen this for `python`.

XzzX:
> > After thinking about it I propose the following structure:
> > on commit (remove on PR as this is already covered by on commit):
> > 
> > * based on ubuntu-latest github actions vm
> > * g++, clang, Intel
> > * debug, release
> > * documentation
> > 
> > weekly, deployment:
> > 
> > * build&push docker containers for all releases
> > * build&push documentation
> > 
> > weekly, compatibility:
> > 
> > * builds for various distros
> > * different MPI versions
> > 
> > I think this is a good compromise between resource consumption and coverage.
> > @junghans Your thoughts?
> 
> Like I said above I don’t like the GitHub action vm, other than that I am ok with the plan. I would maybe push the docker image on every merge to master to avoid confusion.
> 
> However that might mean we will have to split the workflow file as we cannot keep different schedules in one file.

Yes, this would result in three files.

XzzX:
> And the intel compiler package takes long to download and install, so we want to use the container for that.

Github Actions allows to cache steps. But the point you made above is very important. We should "evade" the VM entirely.

junghans:
Ok, what should our default container be? `fedora`?

XzzX:
> Ok, what should our default container be? `fedora`?

I do not have a strong opinion about that. I personally use ubuntu.

.github/workflows/validate.yml
```diff
@@ -11,58 +11,54 @@ jobs:
...
-          python3-h5py
-
-
-
+          mkdir ~/Licenses
+          curl https://dynamicinstaller.intel.com/api/v2/license > ~/Licenses/intel.lic
+      
       - name: Configure
-        run: cmake -S ${{github.workspace}} -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{ matrix.BUILD_TYPE }} -DEXTERNAL_MPI4PY=OFF
+        run: cmake -S . -B build -DCMAKE_BUILD_TYPE=${{ matrix.BUILD_TYPE }} -DCMAKE_CXX_COMPILER=${{ matrix.CXX }} -DEXTERNAL_MPI4PY=${{ matrix.EXTERNAL_MPI4PY }}
```
XzzX, line 45: We will soon also need the `hdf5` library which is a C library. Please also add the C compilers here.

.github/workflows/validate.yml
```diff
@@ -11,58 +11,54 @@ jobs:
     strategy:
       matrix:
         BUILD_TYPE: [Release, Debug]
+        DISTRO: ['fedora', 'fedora_rawhide', 'fedora_mpich', 'opensuse', 'ubuntu', 'ubuntu_devel', 'ubuntu_rolling', 'ubuntu_mpich']
+        CXX: ['g++', 'clang++']
```
XzzX, line 15: Can we get `icpc` also into this list? Can we install the Intel toolsuite for all distributions?

junghans: Somebody could do that, but it won't be me ;-)

.github/workflows/validate.yml
junghans: We can do that when it actually happened. Plus you could write the implement as `extern C`, which would still only require a c++ compiler.

Honestly I would prefer only having one compiler if possible as compiler mixing is another source of errors that can confuse users.

.github/workflows/validate.yml
junghans: spack has support to load modules, but it can also build everything if e.g. the boost has not enable python.

.github/workflows/validate.yml
junghans: https://github.com/votca/csg/blob/master/src/libcsg/modules/io/h5mdtrajectoryreader.cc

junghans:
I will split these in two workflows now, then we can merge this.

junghans:
@XzzX ready for final review

.github/workflows/validate.yml
XzzX: Well, you are only looking for the `CXX` components.
https://github.com/votca/csg/blob/master/src/libcsg/CMakeLists.txt#L18
Interesting this is working since you are using the `C` interface.

If you look for `C` and `HL`, you get the following error:
```
  Unknown extension ".c" for file

   .../espressopp/cmake-build-debug/CMakeFiles/hdf5/cmake_hdf5_test.c

  try_compile() works only for enabled languages.  Currently these are:

    CXX

  See project() command to enable other languages.
```
Can be resolved by adding `C` as a language, however, this requires `C` compilers.

XzzX (approved):
We should remove the pull request trigger. This is already covered by the push trigger and enforced by the branch protection settings.

.github/workflows/validate.yml
junghans: > We should remove the pull request trigger. This is already covered by the push trigger and enforced by the branch protection settings.

Nope it not the same the pull request trigger tests the commit merged with the target branch, while the push trigger just tests the pushed commit itself. I think the former is more important.

junghans:
> We should remove the pull request trigger. This is already covered by the push trigger and enforced by the branch protection settings.

Nope it not the same the pull request trigger tests the commit merged with the target branch, while the push trigger just tests the pushed commit itself. I think the former is more important.

.github/workflows/validate.yml
junghans: I see, we will have to restructure the matix build then.

XzzX:
> Hmm, that is not good!
> 
> We don't have the "branch needs to up to date" option enabled PR and hence without `pull_request:` the merge commit will never be tested.
> 

Ok, I get it. So "branch needs to up to date" xor "pull_request". This boils down to "annoyance to merge multiple times if there are more than one concurrent PR" against "more ci jobs during development". I vote for the second option as long as our build times are pretty fast.

> I would vote for using:
> 
> ```
> push:
>     branches:
>       - master
> ```
> 
> instead

I **strongly** vote against this. I want to have a CI even though I am not ready to submit a pull request for whatever reason. In particular, with the current "best practice" of developing a feature till the end and submitting a LARGE PR this would be devastating as soon as we activate warnings, etc.

XzzX:
Reverted my commit.

XzzX:
Ready to merge from my side.

junghans:
> I **strongly** vote against this. I want to have a CI even though I am not ready to submit a pull request for whatever reason. In particular, with the current "best practice" of developing a feature till the end and submitting a LARGE PR this would be devastating as soon as we activate warnings, etc.

One can always open a draft PR to get CI on every push.

PR #356, opened 2021-04-27
support for parallel H5MD dump and restore

### LINKED ISSUE #346: Check if hdf5 is parallel at CMake time
Currently everything will build, but then the test fails with an error that hdf5 is not parallel

XzzX:
closes #346
fix #356


testsuite/io/test_h5md_parallel.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+#  Copyright (C) 2015-2017(H)
```
junghans, line 2: Huh?

src/io/hdf5.hpp
```diff
@@ -0,0 +1,52 @@
+/*
+  This file is part of ESPResSo++.
```
junghans, line 2: years missing

testsuite/io/test_h5md_parallel.py
XzzX: Copied from one of the existing test cases.

src/io/hdf5.hpp
XzzX: The header is very inconsistent throughout `ESPResSo++`. How do you want it to be?

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,322 @@
+/*
+  This file is part of ESPResSo++.
```
jkrajniak, line 2: Let's stick to this kind of header
```
  Copyright (C) 2021
      Sebastian Eibl

  This file is part of ESPResSo++.
```

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,322 @@
...
+    auto dstSpace = CHECK_HDF5(H5Dget_space(dataset));
+    CHECK_HDF5(H5Sselect_hyperslab(dstSpace, H5S_SELECT_SET, offset.data(), stride.data(),
+                                   count.data(), localDims.data()));
+
+    std::vector<hsize_t> localOffset(globalDims.size(), 0);
+    auto srcSpace = CHECK_HDF5(H5Screate_simple(localDims.size(), localDims.data(), NULL));
+    CHECK_HDF5(H5Sselect_hyperslab(srcSpace, H5S_SELECT_SET, localOffset.data(), stride.data(),
+                                   count.data(), localDims.data()));
+
+    auto datawrite_id = CHECK_HDF5(H5Pcreate(H5P_DATASET_XFER));
```
jkrajniak, line 57: Are there any particular reason to use snake case?

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+    CHECK_HDF5(H5Pclose(datawrite_id));
+    CHECK_HDF5(H5Sclose(dstSpace));
+    CHECK_HDF5(H5Sclose(srcSpace));
+    CHECK_HDF5(H5Dclose(dataset));
+    CHECK_HDF5(H5Sclose(dataspace));
+}
+
+void DumpH5MDParallel::writeHeader(hid_t fileId) {
+    auto group1 = H5Gcreate(fileId, "/h5md", H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
```
jkrajniak, line 69: Shouldn't this be also wrap with `CHECK_HDF5`?

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,322 @@
...
+}
+
+void DumpH5MDParallel::writeBox(hid_t fileId) {
+    std::string groupName = "/particles/atoms/box";
+    auto group = CHECK_HDF5(H5Gcreate(fileId, groupName.c_str(), H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT));
+    std::vector<int> dims = {3};
+    CHECK_HDF5(H5LTset_attribute_int(fileId, groupName.c_str(), "dimension", dims.data(), dims.size()));
+
+    auto boundaryType = H5Tcopy(H5T_C_S1);
+    CHECK_HDF5( H5Tset_size(boundaryType, 8) );
```
jkrajniak, line 95: Unnecessary space
```suggestion
    CHECK_HDF5(H5Tset_size(boundaryType, 8));
```

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,322 @@
...
+    CHECK_HDF5( H5Tset_strpad( boundaryType, H5T_STR_NULLPAD ));
```
jkrajniak, line 96:
```suggestion
    CHECK_HDF5(H5Tset_strpad( boundaryType, H5T_STR_NULLPAD));
```

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,322 @@
...
+    std::vector<hsize_t> boundaryDims = {3};
+    auto space = H5Screate_simple(boundaryDims.size(), boundaryDims.data(), nullptr);
+    auto att = H5Acreate(group, "boundary", boundaryType, space, H5P_DEFAULT, H5P_DEFAULT);
+    CHECK_HDF5(H5Awrite(att, boundaryType, "periodicperiodicperiodic"));
```
jkrajniak, line 100: Why do you have to  put here `periodiperiodicperiodic` ?

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+    std::vector<double> edges = {system_->bc->getBoxL()[0], system_->bc->getBoxL()[1], system_->bc->getBoxL()[2]};
+    CHECK_HDF5(H5LTset_attribute_double(fileId, groupName.c_str(), "edges", edges.data(), edges.size()));
+
+    CHECK_HDF5(H5Gclose(group));
+}
+
+
+void DumpH5MDParallel::writeId(hid_t fileId) {
+    std::string groupName = "/particles/atoms/"+idDataset;
```
jkrajniak, line 113:
```suggestion
    std::string groupName = "/particles/atoms/" + idDataset;
```

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,322 @@
...
+    auto group =
+        H5Gcreate(fileId, groupName.c_str(), H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
+
+    std::vector<int64_t> data;
+    data.reserve(numLocalParticles * 1);
```
jkrajniak, line 118: Do you need this multiplication by `1` ?

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,322 @@
...
+    for (iterator::CellListIterator cit(system_->storage->getRealCells()); !cit.isDone(); ++cit) {
+        data.emplace_back(cit->id());
+    }
+    CHECK_EQUAL(data.size(), numLocalParticles * 1);
```
jkrajniak, line 122: As well here? or this is only for implicit type conversion?

src/io/RestoreH5MDParallel.py
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) 2015-2018
```
jkrajniak, line 1: Please update the header, put your first, last name and institute (if applicable)

jkrajniak (dismissed):

src/io/DumpH5MDParallel.cpp
XzzX: done

src/io/DumpH5MDParallel.cpp
XzzX: no, corrected.

src/io/DumpH5MDParallel.cpp
XzzX: indeed, fixed.

src/io/DumpH5MDParallel.cpp
XzzX: fixed

src/io/DumpH5MDParallel.cpp
XzzX: fixed

src/io/DumpH5MDParallel.cpp
XzzX: fixed.

src/io/RestoreH5MDParallel.py
XzzX: fixed

src/io/DumpH5MDParallel.cpp
XzzX: I am generating all these functions and this is the number of dimensions. I will do this more clearly as I think this also improves readability of the code.

src/io/DumpH5MDParallel.cpp
XzzX: Judging from
https://github.com/espressopp/espressopp/blob/a967dada2b0fda6d390b89abcdc1f1541068d823/src/io/DumpH5MD.py#L216-L227
this seems to be the only supported case.

fedepad (commented):
I made a small comment on the binary compare.

testsuite/io/test_h5md_parallel.py
```diff
@@ -0,0 +1,65 @@
...
+        dump_h5md_parallel.dump()
+
+        self.system.storage.removeAllParticles()
+
+        restore_h5md_parallel = espressopp.io.RestoreH5MDParallel(self.system, 'reference.h5')
+        restore_h5md_parallel.restore()
+        dump_h5md_parallel = espressopp.io.DumpH5MDParallel(self.system, 'dump.h5')
+        dump_h5md_parallel.dump()
+
+        self.assertTrue(binary_compare('reference.h5', 'dump.h5'))
```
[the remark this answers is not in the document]
XzzX: I initially used `h5diff`, however, its a little bit tricky to parse the results correctly. Some errors are returned as return codes, some are printed to `stdout` and some to `stderr`. Therefore, I decided to go with the custom binary compare.

As this is a test case which runs often the file size is kept to a minimum. No parallel comparison is needed here.

testsuite/io/test_h5md_parallel.py
```diff
@@ -0,0 +1,65 @@
...
+
+        ls_dmp = (subprocess.run(['h5ls', '-r', 'reference.h5'], capture_output=True)).stdout
```
junghans, line 57: Oh man, is there no way to do this internally.

Otherwise CMake will need to check for `h5ls`.

XzzX: Sure I can depend on `h5py` or `pyh5md`. I do not want to reimplement a `h5` reader.

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.12)
 
-project(espressopp LANGUAGES CXX)
+project(espressopp LANGUAGES C CXX)
```
junghans, line 3: Instead of adding `C` here, I would add `enable_language(C)` (with a small comment) where `find_package(HFD5)` is called.

src/CMakeLists.txt
```diff
@@ -56,6 +56,10 @@ find_package(Threads REQUIRED)
 set(Boost_USE_MULTITHREADED ON)
 
 add_library(_espressopp ${ESPRESSO_SOURCES})
+#target_link_libraries( _espressopp PRIVATE hdf5::hdf5 hdf5::hdf5_hl ) requires CMake 3.19 :(
```
junghans, line 59: We should consider bumping to 3.19 in the CMake refactor.

src/io/DumpH5MDParallel.cpp
jkrajniak: So this `periodicperiodicperiodic` will be written as three element array?

fedepad (commented):
Since HDF5 is a required dependency by this PR and once merged as seen in the `CMakeLists.txt`, the `README.md` should be updated to include HDF5 as a dependency like the others dependencies that you cite (Boost, etc.).
Unless I am missing it somehow, I don't see the `README.md` in the list of changed files.

CMakeLists.txt
XzzX: What is the benefit of this?

src/CMakeLists.txt
XzzX: In principle I agree. However, currently, CMake 3.19 must be installed explicitly as it is quite new.

XzzX:
> Since HDF5 is a required dependency by this PR and once merged as seen in the `CMakeLists.txt`, the `README.md` should be updated to include HDF5 as a dependency like the others dependencies that you cite (Boost, etc.).
> Unless I am missing it somehow, I don't see the `README.md` in the list of changed files.

`Readme.md` already contains `HDF5 library (>= 1.8) (optional for H5MD file format)`. Currently it is neither looked for nor used, but someone already included it...

src/io/DumpH5MDParallel.cpp
XzzX: yes

CMakeLists.txt
junghans: I think that would make it more clear that the only reason we  need `C` is due to `HDF5`.

testsuite/io/test_h5md_parallel.py
XzzX: I now require `h5py`. I think this is the best way.

CMakeLists.txt
XzzX: ..., `C` is only needed due to a compile test within `FindHDF5.cmake`. The file extension is `c` and, therefore, a c compiler is needed. If we change that to `cpp` we do not need `C` as a language in our project. However, we have to ship our own slightly modified version of `FindHDF5.cmake`. At the same time this would allow us to use exported targets, since we can build upon the CMake 3.20 version. @junghans What do you think?

junghans: I like that! And we should report that issue to cmake upstream as well!

src/io/DumpH5MDParallel.py
```diff
@@ -0,0 +1,57 @@
...
+    def __init__(self, system, filename):
+        cxxinit(self, io_DumpH5MDParallel, system, filename)
+
+    def dump(self):
+        if not (pmi._PMIComm and pmi._PMIComm.isActive() ) or pmi._MPIcomm.rank in pmi._PMIComm.getMPIcpugroup():
+            self.cxxclass.dump(self)
+
+
+
+if pmi.isController:
```
jkrajniak, line 34: Too many lines before

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,479 @@
...
+        .def_readwrite("author", &DumpH5MDParallel::author)
+        .def_readwrite("dumpId", &DumpH5MDParallel::dumpId)
+        .def_readwrite("dumpType", &DumpH5MDParallel::dumpType)
+        .def_readwrite("dumpMass", &DumpH5MDParallel::dumpMass)
+        .def_readwrite("dumpQ", &DumpH5MDParallel::dumpQ)
+        .def_readwrite("dumpGhost", &DumpH5MDParallel::dumpGhost)
+        .def_readwrite("dumpPosition", &DumpH5MDParallel::dumpPosition)
+        .def_readwrite("dumpVelocity", &DumpH5MDParallel::dumpVelocity)
+        .def_readwrite("dumpForce", &DumpH5MDParallel::dumpForce)
+        .def_readwrite("idDataset", &DumpH5MDParallel::idDataset)
```
jkrajniak, line 468: If this writer/restorer is for H5MD format then we shouldn't allow to modify the name of datasets as they are defined in the file format standard. See here: https://nongnu.org/h5md/h5md.html#particles-group
I would remove these properties or keep them as read-only.

doc/sphinx/espressopp.io.DumpH5MDParallel.rst
```diff
@@ -0,0 +1,38 @@
...
+---------------
+
+Dumping a simulation with ``espressopp.io.DumpH5MDParallel``
+
+::
+
+    dump_h5md_parallel = espressopp.io.DumpH5MDParallel(system, 'dump.h5')
+    dump_h5md_parallel.dump()
+
+
```
jkrajniak, line 19: Let's add here 
1. the possible options on what can be stored and how to enable/disable it.
2. the path to the particles location as it is fixed (`/particles/atoms/`)

I would also add *Attention* or *Note* with information that the dump is valid only for V=const (NVT, NVE) simulation as we do not save box dimensions at every simulation step.

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,479 @@
...
+    auto group1 = CHECK_HDF5(H5Gcreate(fileId, "/h5md", H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT));
+    auto group2 =
+        CHECK_HDF5(H5Gcreate(fileId, "/h5md/author", H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT));
+    auto group3 =
+        CHECK_HDF5(H5Gcreate(fileId, "/h5md/creator", H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT));
+    CHECK_HDF5(H5Gclose(group1));
+    CHECK_HDF5(H5Gclose(group2));
+    CHECK_HDF5(H5Gclose(group3));
+
+    std::vector<int> data = {2, 0};
```
jkrajniak, line 88: AFAIK the last version of H5MD is 1.1 (right @pdebuyl ?)

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+    CHECK_HDF5(H5Gclose(group));
+}
+
+void DumpH5MDParallel::writeId(hid_t fileId)
+{
+    using Datatype = int64_t;
+    constexpr int64_t dimensions = 1;  ///< dimensions of the property
+
+    std::string groupName = "/particles/atoms/" + idDataset;
```
jkrajniak, line 133: Let's keep string `/particles/atoms/` as a variable - for safety

src/io/DumpH5MDParallel.cpp
```diff
@@ -0,0 +1,479 @@
...
+    for (iterator::CellListIterator cit(system_->storage->getRealCells()); !cit.isDone(); ++cit)
+    {
+        data.emplace_back(cit->id());
+    }
+    CHECK_EQUAL(int64_c(data.size()), numLocalParticles * dimensions);
+
+    std::vector<hsize_t> localDims = {1, uint64_c(numLocalParticles), dimensions};
+    std::vector<hsize_t> globalDims = {1, uint64_c(numTotalParticles), dimensions};
+
+    std::string dataset_name = "/particles/atoms/" + idDataset + "/value";
```
jkrajniak, line 147: You could use already defind `groupName` variable

doc/sphinx/espressopp.io.DumpH5MDParallel.rst
```diff
@@ -0,0 +1,38 @@
...
+H5MD_ is a format specification on top of the HDF5_ file format.
+
+.. _H5MD: https://nongnu.org/h5md/
+.. _HDF5: https://www.hdfgroup.org/solutions/hdf5/
+
+Dump Simulation
+---------------
+
+Dumping a simulation with ``espressopp.io.DumpH5MDParallel``
```
jkrajniak, line 12: Am I right that this is only to dump a single frame during the simulation?

src/io/DumpH5MDParallel.py
XzzX: Will fix it. At some point we will have a PEP8 formatter :)

src/io/DumpH5MDParallel.cpp
XzzX: fixed

src/io/DumpH5MDParallel.cpp
XzzX: introduced a new variable for 'atoms'

doc/sphinx/espressopp.io.DumpH5MDParallel.rst
XzzX: done

src/CMakeLists.txt
XzzX: I will ship a custom `FindHDF5` until we enforce cmake > 3.19

CMakeLists.txt
XzzX: There was already an attempt to do this:
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4534
TL;DR Enable C language if you use C language bindings.

XzzX:
Ok, since there are no more comments about this merge request we should merge it and revisit the HDF5 output when somebody is actually using it.

@junghans Can you have a final look at it?

junghans (approved):

doc/sphinx/espressopp.io.DumpH5MDParallel.rst
jkrajniak: Ok, so to be clear, the following code:

```python
for step in range(100):
    integrator.run(100)
    h5mdump.dump()
```
will result with 100 frames or only a single last frame? If there will be only a single frame then I would put this information explicitly in documentation. Other dumps like xyz, pdb or gro behave differently, storing multiple frames.

PR #357, opened 2021-04-27
enable make install in ci

### LINKED ISSUE #349: fix install in CMake

XzzX:
closes #349

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -34,8 +34,8 @@ jobs:
       - name: Build
         run: cmake --build ${{github.workspace}}/build --target all -j 2
 
-      #    - name: Install
-      #      run: cmake --install ${{github.workspace}}/build
+      - name: Install
+        run: cmake --install ${{github.workspace}}/build
 
       - name: Test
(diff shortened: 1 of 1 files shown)


junghans (approved):

junghans:
We need to set CMAKE_INSTALL_PATH or use sudo for install

XzzX:
> We need to set CMAKE_INSTALL_PATH or use sudo for install

Why are we not root inside the container?

junghans:
Some tools e.g. MPI don't work as root. Also running as root is really bad practice as the CI won't be able to catch file permission issue.

XzzX:
> Some tools e.g. MPI don't work as root. Also running as root is really bad practice as the CI won't be able to catch file permission issue.

Hmm, all third-party containers I know (not many) do not create user accounts. I think this makes inheriting from it harder. Nevertheless, I agree that, in general, running stuff as root is not optimal. ;) 

Btw. MPI can run as root, `--allow-run-as-root`.

junghans:
Yeah, the user in our containers has `sudo` rights, so one can still become root if needed.

PR #359, opened 2021-04-28
ensure consistent code formatting for the c++ code

### LINKED ISSUE #201: Set travis to check if the code in pull request is formatted correctly

XzzX:
Using `clang-format` to ensure consistent formatting. Run `format.sh` in the root folder to format the whole code base.

closes #202 
closes #201

### DIFF
--- a/.github/workflows/format-command.yml
+++ b/.github/workflows/format-command.yml
@@ -1,0 +1,31 @@
+name: format-command
+on:
+  repository_dispatch:
+    types: [format-command]
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.BOT_PAT }}
+          repository: ${{ github.event.client_payload.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.client_payload.pull_request.head.ref }}
+
+      - name: Format
+        run: ./format.sh
+
+      - name: Commit to the PR branch
+        run: |
+          git config --global user.name 'espressopp-bot'
+          git config --global user.email 'espressopp-bot@no-reply.com'
+          git commit -am "[BOT] clang-format"
+          git push
+
+      - name: Add reaction
+        uses: peter-evans/create-or-update-comment@38214a08ee61b78dc42bd8f45294668c5bfc1be0
+        with:
+          token: ${{ secrets.BOT_PAT }}
+          repository: ${{ github.event.client_payload.github.payload.repository.full_name }}
+          comment-id: ${{ github.event.client_payload.github.payload.comment.id }}
+          reaction-type: hooray
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -153,3 +153,14 @@ jobs:
 
       - name: Upload Report to codecov.io
         uses: codecov/codecov-action@v1
+
+  formatting:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: run formatter
+        run: ./format.sh
+
+      - name: check formatting
+        run: git diff --exit-code
--- a/.clang-format
+++ b/.clang-format
@@ -1,2 +1,7 @@
 BasedOnStyle: Google
 ColumnLimit: 100
+AccessModifierOffset: '-4'
+BinPackParameters: 'false'
+BreakBeforeBraces: Allman
+IndentWidth: '4'
+SortIncludes: false
--- a/.github/workflows/slash-command-dispatch.yml
+++ b/.github/workflows/slash-command-dispatch.yml
@@ -1,0 +1,14 @@
+name: Slash Command Dispatch
+on:
+  issue_comment:
+    types: [created]
+jobs:
+  slashCommandDispatch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Slash Command Dispatch
+        uses: peter-evans/slash-command-dispatch@8e02cf5c38baa3ddce3b993cef6fa49aa33e22ac
+        with:
+          token: ${{ secrets.BOT_PAT }}
+          commands: format
+          repository: espressopp/espressopp
--- a/format.sh
+++ b/format.sh
@@ -1,0 +1,3 @@
+#! /bin/bash
+
+git ls-files -- src*.cpp src*.hpp testsuite*.cpp testsuite*.hpp | xargs clang-format -i -style=file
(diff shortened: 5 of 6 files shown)


junghans:
I would integrate it in CMake instead of using a shell script.

https://github.com/ECP-copa/Cabana/blob/master/CMakeLists.txt#L205

XzzX:
> I would integrate it in CMake instead of using a shell script.
> 
> https://github.com/ECP-copa/Cabana/blob/master/CMakeLists.txt#L205

Make targets are only available in the build folder. Why should I go to the build folder to format my source folder? I consider the wild mix of source and build folders in `espressopp` bad practice.

junghans:
True, but we need a way to check the clang-format version, otherwise it will be a constant back and forth between version.

plus CMake support Windows ;-)

.github/workflows/validate.yml
```diff
@@ -129,3 +129,14 @@ jobs:
       - name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
 
+  formatting:
+    runs-on: ubuntu-latest
```
junghans, line 133: I would run that in one of our containers as `ubuntu-latest` has 3 different clang versions.

junghans:
We should also run pep8 on all the python!

XzzX:
> True, but we need a way to check the clang-format version, otherwise it will be a constant back and forth between version.

Hmm, enforcing a specific `clang-format` version seems awkward. Which version would you use? Default in Ubuntu 18, 20, fedora? We also have to force the users to have the same version... Or we hope that they all produce the same results. I formatted with `clang-format-12`, seems to be consistent with what github actions provides.

> 
> plus CMake support Windows ;-)

True, have you ever tested `espressopp` on Windows? If not I have serious doubts it will compile. My experience with `waLBerla` (we support Windows) is that MSVC is tricky at times. ;) And there is only support for OpenMP 2.0...

XzzX:
> We should also run pep8 on all the python!

On step after the other :) Let us see what the others say in today's meeting.

.github/workflows/validate.yml
XzzX: Which one?

XzzX:
Different versions of `clang-format` seem to be an issue.
https://stackoverflow.com/questions/59395507/how-to-maintain-a-clang-format-file-for-different-clang-format-versions
Ideas? Force a certain version?

junghans:
In VOTCA we use the version that comes with fedora:latest. Ubuntu would be ok as well, that might have the advantage that there is only a new lts release every couple of years.

XzzX:
> In VOTCA we use the version that comes with fedora:latest. Ubuntu would be ok as well, that might have the advantage that there is only a new lts release every couple of years.

Ok, we can try it like this. If troubles arise we can look at it again.
I am not really troubled with updates of the `clang-format` version within a distribution, more with users having multiple different installations...

XzzX:
Since this change influences everyone, who should give their ok?

junghans:
In VOTCA we also have an action that allow formatting of a PR (see https://github.com/votca/tools/blob/master/.github/workflows/format.yml), we could to the same for espressopp to help developers in case they cannot get the right version of `clang-format`.

XzzX:
> In VOTCA we also have an action that allow formatting of a PR (see https://github.com/votca/tools/blob/master/.github/workflows/format.yml), we could to the same for espressopp to help developers in case they cannot get the right version of `clang-format`.

This is an option, yes, at the cost of adding additional commits.

XzzX:
> In VOTCA we also have an action that allow formatting of a PR (see https://github.com/votca/tools/blob/master/.github/workflows/format.yml), we could to the same for espressopp to help developers in case they cannot get the right version of `clang-format`.

I see another problem with this approach. This does not stop multiple changes in PRs due to formatting that will get reverted afterwards. This makes reviewing more complicated.

junghans:
> > In VOTCA we also have an action that allow formatting of a PR (see https://github.com/votca/tools/blob/master/.github/workflows/format.yml), we could to the same for espressopp to help developers in case they cannot get the right version of `clang-format`.
> 
> I see another problem with this approach. This does not stop multiple changes in PRs due to formatting that will get reverted afterwards. This makes reviewing more complicated.

Yeah it only makes sense to the formatting once at the end or simply do a squash merge.

junghans:
I just realized there is a `format` target already in CMake.

XzzX:
> I just realized there is a `format` target already in CMake.
> 
> https://github.com/espressopp/espressopp/blob/5df088dcc353cad0c472e0ac1965e4dfb2cb6acd/src/CMakeLists.txt#L96-L105

This PR would remove this target.

XzzX:
My current suggestion:

1) Convert bash script to Python. This should allow potential Windows users to also run the script.
2) Run format check per commit but not as required. Reviewers can then manually recommend running the script to users.

I am a little bit hesitant to include a format&commit bot as I am afraid it will make change history less useful.

@junghans @jnvance Opinions?

junghans:
From my experience on other projects even with a pre-commit hook basically everybody forgets to format and then everything is reformated in the last commit. There is a clang tool to run formatting on each commit, but that can generate a lot of conflicts.

I thought about this quite a bit but I don't have a good solution.

Looking at the most recent PRs our history is pretty useless already, so one more formatting commit won't make it much worse. Maybe we should just squash merge PRs by default, then formatting commits would not matter.

XzzX:
> From my experience on other projects even with a pre-commit hook basically everybody forgets to format and then everything is reformated in the last commit. 

Use an IDE that supports format on save :)

> Maybe we should just squash merge PRs by default, then formatting commits would not matter.

That is a very good suggestion. I like it. 

If I understand the VOTCA bot you can call him via the comment function and he produces a format commit? Are the custom actions and votca bot user still required to implement this functionality?

junghans:
The only reason we need a bot account is that only a commit from a PAT can trigger actions and usually you want to run the CI again after formatting.

junghans (approved):

.github/workflows/format-command.yml
```diff
@@ -0,0 +1,31 @@
+name: format-command
+on:
+  repository_dispatch:
+    types: [format-command]
+jobs:
+  example:
```
junghans, line 6: Might be not a great name.

junghans (changes requested):

.github/workflows/format-command.yml
XzzX: changed to format

junghans (approved):

PR #362, opened 2021-04-28
CI: auto-cancel old jobs

junghans:


XzzX:
I need to have a detailed look at `potiuk/cancel-workflow-runs@master`.

XzzX (approved):
seems useful

PR #364, opened 2021-04-30
replaced icpc build with oneapi release build

XzzX:
For the time being removed fedora icpc builds. Since we still have Intel builds I think this is ok.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -15,9 +15,9 @@ jobs:
         CXX: ['g++', 'clang++']
         EXTERNAL_MPI4PY: ['On']
         include:
-          - BUILD_TYPE: 'Debug'
-            DISTRO: 'fedora_intel'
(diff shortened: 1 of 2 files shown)


XzzX:
@junghans Is the deviation observed in the test case acceptable?

junghans:
`icpx` is totally different llvm-based compiler, I would just disable the debug builds for the time being.

XzzX:
> `icpx` is totally different llvm-based compiler, I would just disable the debug builds for the time being.

Also fine with me, just create a working setup. 

In my opinion, in the current state enabling compiler warnings would be much more urgent and helpful than what compilers we are using.

junghans:
I think `icpc` should be fixed now.

XzzX:
Let's see. Rerunning master.

XzzX:
Obsolete. Changes in `buildenv` solved the issue.

PR #367, opened 2021-05-03
Re-enable coverage reporting

### LINKED ISSUE #348: re-enable codecov in CI

junghans:
Fix #348

Coverage for:
- [x] C++
- [x] Python


junghans:
@XzzX @jkrajniak any input on python coverage, I know `coverage run -m pytest`, but we aren't using `pytest` here.

XzzX:
> @XzzX @jkrajniak any input on python coverage, I know `coverage run -m pytest`, but we aren't using `pytest` here.

I have no idea, would have to do research myself.

.github/workflows/validate.yml
```diff
@@ -133,3 +133,20 @@ jobs:
...
+  coverage:
+    runs-on: ubuntu-latest
+    container: espressopp/buildenv:fedora
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Coverage build
+        run: |
+          cmake -B builddir -DUSE_GCOV=on -DCMAKE_BUILD_TYPE=Debug
+          cmake --build builddir --verbose -j 2
```
XzzX, line 145: general comment: Is the verbose output really beneficial? I think all the include paths etc hide the important information.

fedepad:
> @XzzX @jkrajniak any input on python coverage, I know `coverage run -m pytest`, but we aren't using `pytest` here.

similar command depending on the test module used. E.g. if unittests is used for the tests, it is something like: `coverage run -m unittest discover` to be adapted to the specific case.
https://coverage.readthedocs.io/en/coverage-5.5/index.html?highlight=unittest#quick-start
Would be interesting to see if any usage of either subprocess or some threads calls if any that might eventually give some "troubles" (https://coverage.readthedocs.io/en/coverage-5.5/trouble.html#things-that-don-t-work) but you will only know when you try it.

.github/workflows/validate.yml
junghans: I mainly wanted to make sure there is actually a `--coverage` flag.

I personally like to have verbose output so that all the information is there for debugging.

junghans:
> > @XzzX @jkrajniak any input on python coverage, I know `coverage run -m pytest`, but we aren't using `pytest` here.
> 
> similar command depending on the test module used. E.g. if unittests is used for the tests, it is something like: `coverage run -m unittest discover` to be adapted to the specific case.
> https://coverage.readthedocs.io/en/coverage-5.5/index.html?highlight=unittest#quick-start
> Would be interesting to see if any usage of either subprocess or some threads calls if any that might eventually give some "troubles" (https://coverage.readthedocs.io/en/coverage-5.5/trouble.html#things-that-don-t-work) but you will only know when you try it.

Good point I think we use `unittest` in the `testsutie` so that will help.

XzzX (commented):
`codecov/codecov-action@v1` finds both `g++` and `python` code coverage files automatically?

Please add the code coverage badge to `README.md` again.

How is code coverage reported? Will this chart be pushed into the conversation after every commit? Will there be multiple charts within the conversation?

junghans:
> `codecov/codecov-action@v1` finds both `g++` and `python` code coverage files automatically?

Yes, one can see that in logfile, I am still a bit puzzled about the python coverage.
> 
> Please add the code coverage badge to `README.md` again.


will do
> 
> How is code coverage reported? Will this chart be pushed into the conversation after every commit? Will there be multiple charts within the conversation?

There is one comment per PR and a status. One can also look details up on the codecov.io website.

junghans:
> > Please add the code coverage badge to `README.md` again.
> 
> will do
> 
Done

XzzX:
> > `codecov/codecov-action@v1` finds both `g++` and `python` code coverage files automatically?
> 
> Yes, one can see that in logfile, I am still a bit puzzled about the python coverage.

Is a coverage of around 50% realistic? Or do we just run some complete simulations without any checks to get the number high? :)

XzzX (approved):

junghans:
> > > `codecov/codecov-action@v1` finds both `g++` and `python` code coverage files automatically?
> > 
> > 
> > Yes, one can see that in logfile, I am still a bit puzzled about the python coverage.
> 
> Is a coverage of around 50% realistic? Or do we just run some complete simulations without any checks to get the number high? :)

Coverage only means the code was ran, it doesn't check if garbage comes out though.

XzzX:
> > > > `codecov/codecov-action@v1` finds both `g++` and `python` code coverage files automatically?
> > > 
> > > 
> > > Yes, one can see that in logfile, I am still a bit puzzled about the python coverage.
> > 
> > 
> > Is a coverage of around 50% realistic? Or do we just run some complete simulations without any checks to get the number high? :)
> 
> Coverage only means the code was ran, it doesn't check if garbage comes out though.

Therefore, we should not run anything that does not check for garbage. Otherwise, the number is pretty useless.

PR #370, opened 2021-05-05
cmake options for vectorization report and instruction set

XzzX:


CMakeLists.txt
```diff
@@ -22,6 +22,8 @@ set(CMAKE_CXX_EXTENSIONS OFF) #...without compiler extensions like gnu++11
 ########################################################################
 # User input options                                                   #
 ########################################################################
+option(ESPP_LOCAL_ARCHITECTURE "Use instruction set of the local architecture." OFF)
+option(ESPP_VEC_REPORT "Enable reporting of loop vectorization." OFF)
```
junghans, line 26:
```suggestion
option(ESPP_VEC_REPORT "Enable reporting of loop vectorization (only for Intel compiler)." OFF)
```

src/CMakeLists.txt
```diff
@@ -56,18 +56,41 @@ find_package(Threads REQUIRED)
...
+    endif()
+endif()
+if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
+    if(ESPP_LOCAL_ARCHITECTURE)
+        target_compile_options(_espressopp PRIVATE "-march=native")
+    endif()
+endif()
+if(CMAKE_CXX_COMPILER_ID MATCHES Intel)
+    if(ESPP_LOCAL_ARCHITECTURE)
+        target_compile_options(_espressopp PRIVATE "-xHost")
```
jnvance, line 73: Can you also add `-restrict` here?

jnvance (changes requested):

junghans (approved):

src/CMakeLists.txt
XzzX: Interesting. Is my understanding correct that Intel uses this flag to enable checking for the `restrict` keyword? I hope it does not transform all pointers into `restrict`. And it only applies to `restrict`, `__restrict` is always enabled? Pretty weird,
https://software.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/compiler-options/compiler-option-details/language-options/restrict-qrestrict.html

src/CMakeLists.txt
```diff
@@ -56,18 +56,41 @@ find_package(Threads REQUIRED)
 set(Boost_USE_MULTITHREADED ON)
 
 add_library(_espressopp ${ESPRESSO_SOURCES})
+
+# compiler specific configurations
+if(CMAKE_CXX_COMPILER_ID MATCHES GNU)
+    if(ESPP_LOCAL_ARCHITECTURE)
+        target_compile_options(_espressopp PRIVATE "-march=native")
```
junghans, line 63: Will it lead to problems if only the library is compiled with optimization flags, but not the executables (tests)?

XzzX: No, this does not change the ABI. It only tells the compiler which instruction set it is allowed to used.

src/CMakeLists.txt
XzzX: added

jnvance: I think it's the latter. I use a lot of them in the pointer declarations in the vec submodule.

XzzX: My state is, I do think we do not need it but it does not hurt. So we can enable it.

jnvance (approved):

src/CMakeLists.txt
jnvance: There is even `-qopt-zmm-usage=high` recommended for AVX-512 which slightly improves loop vectorization. But we can also decide to add it later.

PR #372, opened 2021-05-05
Enable *.py tests

jnvance:
Tests were not parsed correctly:
https://github.com/espressopp/espressopp/blob/6c02d98fc2af3ed999a8954d79a10ed4110d6caa/testsuite/CMakeLists.txt#L31-L36

Enabled tests in directories with only *.py files that were not executed and fixed some unit tests that were failing.


junghans (changes requested):
I saw this issue in #367, too, could you also remove all now redundant `testsuite/*/CMakeLists.txt` ?

jnvance:
> I saw this issue in #367, too, could you also remove all now redundant `testsuite/*/CMakeLists.txt` ?

Yup. Also, there should now be 80 tests from 66 previously.

testsuite/RealNDTest/CMakeLists.txt
```diff
@@ -1,5 +0,0 @@
-include_directories(${CMAKE_CURRENT_SOURCE_DIR}/../../src/include)
-add_definitions (-DBOOST_TEST_DYN_LINK)
-add_executable(RealNDTest RealNDTest.cpp)
```
junghans, line 3: Not sure we can remove this file as it is non-standard.

testsuite/pickle_potential/CMakeLists.txt
```diff
@@ -1,4 +0,0 @@
-add_test(pickle_potential ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/pickle_potential.py)
```
junghans: Not sure we can remove this is standard, not starting with `Test*`.

testsuite/pickle_potential/CMakeLists.txt
```diff
@@ -1,4 +0,0 @@
...
-set_tests_properties(pickle_potential PROPERTIES ENVIRONMENT "${TEST_ENV}")
-add_test(testwarmup ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/testwarmup.py)
```
junghans: Same here

junghans: OK, you moved it.

testsuite/RealNDTest/CMakeLists.txt
XzzX: The include should be irrelevant as nothing is included from `src/include`. Therefore, it should be "standard".

XzzX:
If we start renaming test files then please stick to the naming convention of `unittest`.
https://docs.python.org/3/library/unittest.html#unittest-test-discovery
This means `test*.py`. Then we can also use discovery.
Also `pytest` will then be able to recognize our tests.
I vote against using the `-p` flag and doing some custom naming.

jnvance:
Hi @songbin6280, could you verify the numbers in TestSmoothSquareWell
https://github.com/espressopp/espressopp/blob/ad52818fcefb5149c6564bf5d2e4b1fac4a3fb0b/testsuite/interaction_potentials/TestSmoothSquareWell.cpp#L68
Error: https://github.com/espressopp/espressopp/pull/372/checks?check_run_id=2513487629#step:9:56

The test was not running before so this error was only caught now.

testsuite/RealNDTest/CMakeLists.txt
junghans: Ok, got it, the blob should work for that test as well

testsuite/interaction_potentials/unittest/testFENE.py
```diff
@@ -20,11 +20,11 @@
 
 
 import unittest
-import espressopp.unittest
+import espressopp
 from espressopp.interaction.FENE import *
 from espressopp import Real3D, infinity
 
-class Test0FENE(espressopp.unittest.TestCase) :
+class Test0FENE(espressopp.tools.TestCase) :
```
XzzX, line 27: Is this test executed? Does not seem like that. Also `espressopp.tools.TestCase` does not exist?

XzzX: Occurs on multiple locations.

XzzX (changes requested):

testsuite/interaction_potentials/unittest/testFENE.py
jnvance: Yup, this is executed. There are now 82 tests from the 66 in the master branch. Some were not executed previously. See: https://github.com/espressopp/espressopp/pull/372/checks?check_run_id=2519167875#step:9:54

I relocated the unittest module to tools since it was not being imported inside the testsuite directory. https://github.com/jnvance/espressopp/blob/ba953bd922fd78e3cf1c1cbb58f94f71d3e7d9e2/src/tools/test.py

XzzX: Ah, I missed the file you moved. Perfect.

I do not like that `espressopp.tools.TestCase` is not adhering to `unittest` naming, but that's a different issue.

XzzX (approved):

jnvance:
@junghans could you review this as well. Merging is blocked since you requested changes and there's no resolve button.

junghans (approved):

songbin6280:
> Hi @songbin6280, could you verify the numbers in TestSmoothSquareWell
> 
> https://github.com/espressopp/espressopp/blob/ad52818fcefb5149c6564bf5d2e4b1fac4a3fb0b/testsuite/interaction_potentials/TestSmoothSquareWell.cpp#L68
> 
> Error: https://github.com/espressopp/espressopp/pull/372/checks?check_run_id=2513487629#step:9:56
> 
> The test was not running before so this error was only caught now.
> 
> PS: I've temporarily ignored the test.

Thank you for letting me know. I created a pull request #393, in which I fixed the values and left the ignore file untouched.

Issue #351, opened 2021-04-26
drop internal mpi4py? 


junghans:

XzzX:
https://github.com/espressopp/espressopp/blob/a967dada2b0fda6d390b89abcdc1f1541068d823/src/System.cpp#L52-74

This is the only reference of `mpi4py` in the C++ code. Coupling would be much cleaner if just `MPI_Comm` was passed. However, I have no clue how `mpi` is wrapped through `mpi4py` is wrapped through `pmi` and passed to the backend.

@stuehn Can we couple with `MPI_Comm` here?

After refactoring this part I am ok with dropping our custom `mpi4py`.

junghans:
Everything seems to work with v2 and v3, so couldn't we just check the detected version in CMake?

junghans:
Seeing the most recent problems with our internal mpi4py, I think we need a CI container without mpi4py to test this more in depth.

XzzX:
We should couple via a native MPI type and not via some mpi4py struct. This way our C++ code would be independent of mpi4py.
see #396

junghans:
After #396 is merged, can we now drop the internal `mpi4py` and install it via `requirements.txt`?

XzzX:
Yes, I think so. I will provide a PR.

PR #374, opened 2021-05-06
modernize cmake

### LINKED ISSUE #342: cleanup cmake and adhere to modern cmake

XzzX:
closes #342


src/CMakeLists.txt
```diff
@@ -80,9 +77,20 @@ endif()
 get_target_property(ESPP_CFLAGS _espressopp COMPILE_OPTIONS)
 message(STATUS "_espressopp compiler flags are: ${ESPP_CFLAGS}")
 
-target_link_libraries(_espressopp ${Boost_LIBRARIES} ${PYTHON_LIBRARIES} MPI::MPI_CXX ${FFTW3_LIBRARIES} ${VAMPIRTRACE_LIBRARIES})
+target_include_directories(_espressopp PUBLIC ${CMAKE_SOURCE_DIR}/src)
```
junghans, line 80: All these 3 should be marked e.g. `$<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/src>` as they aren't useful after install.

src/CMakeLists.txt
```diff
@@ -80,9 +77,20 @@ endif()
...
+target_include_directories(_espressopp PUBLIC ${CMAKE_SOURCE_DIR}/src/include)
+target_include_directories(_espressopp PUBLIC ${CMAKE_BINARY_DIR}/src)
+target_include_directories(_espressopp PRIVATE ${FFTW3_INCLUDES})
```
junghans, line 83: We should create an imported target in our custom `FindFFTW3.cmake`.

See  https://github.com/votca/tools/blob/master/CMakeModules/FindFFTW3.cmake#L40-L51

CMakeLists.txt
```diff
@@ -56,16 +56,13 @@ find_package(MPI REQUIRED COMPONENTS CXX)
...
 
 find_package(FFTW3 REQUIRED)
-include_directories(${FFTW3_INCLUDES})
 
 ########################################################################
 #Process GROMACS settings
 ########################################################################
 
 if(WITH_XTC)
     find_package(GROMACS REQUIRED)
```
junghans, line 65: We should install this library in our CI and delete our custom find module as gromacs brings it's own now.

junghans: I.e. `find_package(GROMACS 2021 CONFIG NAMES gromacs gromacs_d)`

src/CMakeLists.txt
XzzX: Also true for `mpi4py`. Anyone volunteering? :)

CMakeLists.txt
XzzX: Yes.

General remark. As far as I can see Gromacs is only used to write `XTC` files. Perhaps it is a little bit of an overkill to require a Gromacs installation just for that...

junghans: It is optional after all.

XzzX: True. Is Gromacs installed in our buildenv?

src/CMakeLists.txt
junghans: I would leave that until #351 is done.

src/CMakeLists.txt
```diff
@@ -80,9 +58,19 @@ endif()
 get_target_property(ESPP_CFLAGS _espressopp COMPILE_OPTIONS)
 message(STATUS "_espressopp compiler flags are: ${ESPP_CFLAGS}")
 
-target_link_libraries(_espressopp ${Boost_LIBRARIES} ${PYTHON_LIBRARIES} MPI::MPI_CXX ${FFTW3_LIBRARIES} ${VAMPIRTRACE_LIBRARIES})
+target_include_directories(_espressopp PUBLIC $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/src> $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/src/include>
+     $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/src>)
```
XzzX, line 62: Do we also need `INSTALL_INTERFACE` for the installed version to work?

contrib/mpi4py/CMakeLists.txt
```diff
@@ -4,14 +4,14 @@ set(MPI4PY_DIR mpi4py-3.0.3)
 set(MPI4PY_SOURCES ${MPI4PY_DIR}/src/MPI.c)
 set(DL_SOURCES ${MPI4PY_DIR}/src/dynload.c)
 
-include_directories(${CMAKE_CURRENT_SOURCE_DIR}/${MPI4PY_DIR}/src)
-
 add_library(MPI ${MPI4PY_SOURCES})
-target_link_libraries(MPI ${Python3_LIBRARIES} ${VAMPIRTRACE_LIBRARIES} MPI::MPI_CXX)
+target_link_libraries(MPI PUBLIC Python3::Python MPI::MPI_CXX)
+target_include_directories(MPI PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/${MPI4PY_DIR}/src)
```
XzzX, line 9: Do we need `BUILD_INTERFACE` here as well?

XzzX (commented):
Thank you @junghans. We are getting somewhere :)

src/CMakeLists.txt
junghans: We don't install any headers, so there is no install interface ;-)

contrib/mpi4py/CMakeLists.txt
junghans: Good catch!

src/CMakeLists.txt
XzzX: Ok, so we do not use the raw usage of `_espresso.so` without the Python layer? @stuehn mentioned it once. We do not have raw C++ users?

junghans:
I think this ready.

XzzX:
Fine from my side.

src/CMakeLists.txt
junghans: We can always add an install interface once we have users for it.

PR #377, opened 2021-05-07
rely on requirements.txt for python dependencies

XzzX:


fedepad (commented):
Are there any specific version requirements for the python packages or it is fine as is?

XzzX:
> Are there any specific version requirements for the python packages or it is fine as is?

Good point. I do not know. Perhaps someone with more experience with the framework can tell.

junghans:
It would be nice to know how long the installation with pip takes once we purge stuff from the containers.

XzzX:
> It would be nice to know how long the installation with pip takes once we purge stuff from the containers.

What do you suggest apart from trying it out? I do not think it will take too long. I think it will download binary wheels.

junghans (approved):

PR #383, opened 2021-05-10
Speedup polymer_melt/halfcell tests

### LINKED ISSUE #382: polymer_melt_tabulated_thirdcell is not a test...
`82/82 Test #80: polymer_melt_tabulated_thirdcell .....   Passed  3103.93 sec`

This test runs for 51 minutes in debug mode. That is not a test. @jnvance Can we make this faster or move it to exampeles?

jnvance:
Fixes #382


junghans:
@ppkk any comments?

junghans:
> @junghans what do you mean with different code path?  The halfCell, third... and so on functionality can be "activated" via the halfCellInt FLAG, or did I miss some other code paths... 

How many different options/flags are there? 2 or 3?

govarguz:
In theory, you could split the cell by any integer. However, only in very special cases halfCellInt values > 2 (Halfs) will be used, because only a few systems will profit from it....as shown for the polymer melt test with such small cutoffs where performance worsen. As far as there is one test for halfCells and the documentation describes properly how halfCellInt works for advanced users, I think no thridCells test is strictly required. Cheers

junghans (approved):

PR #387, opened 2021-05-10
enable Werror in CI

XzzX:
We need a solution for
https://github.com/boostorg/python/issues/359

It seems like warnings from within Boost::Python are not correctly suppressed. Most probably due to inlining.


src/CMakeLists.txt
```diff
@@ -40,11 +40,17 @@ if(CMAKE_CXX_COMPILER_ID MATCHES GNU)
     if(ESPP_LOCAL_ARCHITECTURE)
         target_compile_options(_espressopp PRIVATE "-march=native")
     endif()
+    if(ESPP_WERROR)
+        target_compile_options(_espressopp PRIVATE "-Werror")
+    endif()
```
junghans, line 45: As it is the same block everywhere, could we just write it one time on the outside.

XzzX: I also thought about this. I do not know if we want to add another compiler in the future. The current way is clearly structured but verbose and a little bit of duplication. I can live with both.

PR #389, opened 2021-05-14
fix warnings and enable -Wall

XzzX:


src/CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ if(ESPP_WERROR)
 endif()
 # compiler specific configurations
 if(CMAKE_CXX_COMPILER_ID MATCHES GNU)
```
junghans, line 42: Why are we only doing this for GCC? Or is clang and intel for later?
We should add `-Wextra` and `-pedantic` later, too.
And why the `-Wno-unknown-pragmas`?

Also there should be an option to disable it.

XzzX: > Why are we only doing this for GCC? Or is clang and intel for later?
> We should add `-Wextra` and `-pedantic` later, too.

Clang produces a lot more warnings. I spent half a day on fixing the warnings to get g++ working. I agree we should add more warnings but I will not do it alone. Futhermore, at some places the warnings hint that the current code construct is error prone. So just fixing the warning does not eliminate the underlying cause. However, again, I cannot do this alone.

> And why the `-Wno-unknown-pragmas`?

Due to Intel specific vectorization pragmas.

> 
> Also there should be an option to disable it.

NO, either we build warning free or we do not. Someone writing code that produces warnings and somebody else who fixes them is not how it works.

junghans: > > Why are we only doing this for GCC? Or is clang and intel for later?
> > We should add `-Wextra` and `-pedantic` later, too.
> 
> Clang produces a lot more warnings. I spent half a day on fixing the warnings to get g++ working. I agree we should add more warnings but I will not do it alone. Futhermore, at some places the warnings hint that the current code construct is error prone. So just fixing the warning does not eliminate the underlying cause. However, again, I cannot do this alone.
> 
Ok we should work out way up later then.

> > And why the `-Wno-unknown-pragmas`?
> 
> Due to Intel specific vectorization pragmas.

We should hide these inside `#ifdef __INTEL_COMPILER` then.

> 
> > Also there should be an option to disable it.
> 
> NO, either we build warning free or we do not. Someone writing code that produces warnings and somebody else who fixes them is not how it works.

I disagree, warnings should be one by default, yes, but future versions of compilers will warn about different things, which can be quite annoying, hence there should be a switch to disable `-Wall`.

junghans (approved):

PR #390, opened 2021-05-14
Fix usage of CMAKE_MODULE_PATH

jnvance:
CMAKE_MODULE_PATH is a list of directories so it should not be used to locate cmake/gitversion.cmake.

Found this while trying to use an external library that appends to CMAKE_MODULE_PATH.


junghans (approved):

XzzX (approved):
I also fixed that in one of my branches.

PR #393, opened 2021-05-17
fix: more precision in SmoothSquareWell potential test

songbin6280:
Hello, 

Based on the comments here: https://github.com/espressopp/espressopp/pull/372, I added more precision to the values to the test of smooth square well potential. I ran the test, and it passed on my machine. I did NOT remove ignore file in this pull request.

Thank you,
Bin


jnvance:
Thanks for the fix. I deleted IGNORE so we can see if the test runs successfully on the CI.

jnvance (approved):

PR #394, opened 2021-05-17
cmake: fix mpi4py include path

junghans:
Next attempt to fix `weekly` build.

The docker build in the CI doesn't fail because we are using the mpi4py system header even with `EXTERNAL_MPI4PY=ON` (on Ubuntu). On Fedora the headers are in a non-standard location hence the failure.

### DIFF
--- a/contrib/mpi4py/CMakeLists.txt
+++ b/contrib/mpi4py/CMakeLists.txt
@@ -16,7 +16,7 @@ if (NOT ${CMAKE_CURRENT_SOURCE_DIR} STREQUAL ${CMAKE_CURRENT_BINARY_DIR})
 endif()
 
 add_library(MPI4PY_include INTERFACE)
-target_include_directories(MPI4PY_include INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/${MPI4PY_DIR}/src/mpi4py/include>)
+target_include_directories(MPI4PY_include SYSTEM INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/${MPI4PY_DIR}/src/mpi4py/include>)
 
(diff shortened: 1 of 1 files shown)


XzzX (approved):

junghans:
@XzzX Oh no, now we have a warning in an external header:
```
#10 112.3 /home/espressopp/espressopp/contrib/mpi4py/mpi4py-3.0.3/src/mpi4py/include/mpi4py/mpi4py.h:16:12: error: 'int import_mpi4py()' defined but not used [-Werror=unused-function]
#10 112.3    16 | static int import_mpi4py(void) {
#10 112.3       |            ^~~~~~~~~~~~~
#10 114.2 cc1plus: all warnings being treated as errors
#10 114.2 make[2]: *** [src/CMakeFiles/_espressopp.dir/build.make:336: src/CMakeFiles/_espressopp.dir/System.cpp.o] Error 1
```

XzzX:
> @XzzX Oh no, now we have a warning in an external header:
> 
> ```
> #10 112.3 /home/espressopp/espressopp/contrib/mpi4py/mpi4py-3.0.3/src/mpi4py/include/mpi4py/mpi4py.h:16:12: error: 'int import_mpi4py()' defined but not used [-Werror=unused-function]
> #10 112.3    16 | static int import_mpi4py(void) {
> #10 112.3       |            ^~~~~~~~~~~~~
> #10 114.2 cc1plus: all warnings being treated as errors
> #10 114.2 make[2]: *** [src/CMakeFiles/_espressopp.dir/build.make:336: src/CMakeFiles/_espressopp.dir/System.cpp.o] Error 1
> ```

Hmmm, in principle this should not be an issue.
1) Neither our `-Wall` nor our `-Werror` should propagate to the external library since they are both private?!?
2) External libraries should be marked `system` and therefore should not display any warning...

junghans:
Let's see if that helps.

PR #395, opened 2021-05-17
cmake: install in ~/.local by default

junghans:


XzzX:
Why not `cmake --install <dir> --prefix=...`. This seems to be the cmake standard to install into a custom location.
https://cmake.org/cmake/help/v3.20/manual/cmake.1.html#install-a-project

In general I do not like to change common defaults. It seems odd to default install into home and not `/usr/local`.

junghans:
> Why not `cmake --install <dir> --prefix=...`. This seems to be the cmake standard to install into a custom location.
> https://cmake.org/cmake/help/v3.20/manual/cmake.1.html#install-a-project

The `--prefix` option only works for relocateable packages, which esp++ is not due `ESPRC` (and maybe others).
I.e. if one installs with `cmake --install` and a prefix different from `CMAKE_INSTALL_PREFIX`, `CMAKE_INSTALL_PREFIX` will still show up in ESPRC.

> 
> In general I do not like to change common defaults. It seems odd to default install into home and not `/usr/local`.

Well, we also changed the defaults of `BUILD_SHARED_LIBS`, `CMAKE_BUILD_TYPE` and a couple of other CMake default settings. Ultimately, I think the case of a user installing esp++ in `/usr/local/` is very rare and hence `~/.local` seems a more reasonable default for us.

XzzX:
Then let's install directly into `site-packages` and get rid of ESPRC.

junghans:
> Then let's install directly into `site-packages` and get rid of ESPRC.

which site-packages? The one in `~/.local`?

junghans:
I meant `~/.local/lib/python*/site-packages`.

XzzX:
> I meant `~/.local/lib/python*/site-packages`.

Yes. This is how `pip` would install it. I just see some caveats.

1) With `pip` you could do a development install which links the files instead of copying them. This way you could recompile and have the newest version immediately. Now you have to reinstall.
2) `pip` installs some description files together with the package. We currently do not do that.
3) `pip` checks global `\usr\local' first and falls back to `.local` if it is not writable. With the proposed setup we "cannot" install globally.
4) I do not know if our "detect install directory" works well with python environments or conda.

junghans:
This PR would set the default for main prefix to `~/.local` and `lib/python*/site-packages` gets appended by python resulting in `~/.local/lib/python*/site-packages`.

XzzX:
@junghans What do you think about something build on this https://github.com/espressopp/espressopp/pull/401

junghans:
> @junghans What do you think about something build on this #401 

I would like to keep the cmake install for spack etc.

PR #398, opened 2021-05-18
cmake: remove ESPRC

junghans:
https://github.com/espressopp/espressopp/pull/395#issuecomment-843265788


README.md
```diff
@@ -46,12 +46,11 @@ ESPResSo++ requires Python 3.7 or newer. All required Python packages are listed
 $ cd espressopp
 $ cmake .
 $ make -j2
```
jnvance, line 48: Should `make install` also be called?

junghans: It was not mentioned there before...maybe something we should add in #395.

This part needs a clean-up, too, as in-source builds are really not a thing anymore.

XzzX (commented):
`doc/sphinx/installation.rst` is partly a duplication of `README.md`. There should only be one source of truth. I vote for removing all installation stuff from the README and just keep the docker quickstart in the README. For installation link to the documentation.

jnvance:
The changes LGTM. However, there are currently no instructions to be able to use espp outside of the build directory (e.g. appending build dir to `PYTHONPATH`, setting -D CMAKE_INSTALL_PREFIX or --prefix, make install, etc).

jnvance (approved):

junghans:
> `doc/sphinx/installation.rst` is partly a duplication of `README.md`. There should only be one source of truth. I vote for removing all installation stuff from the README and just keep the docker quickstart in the README. For installation link to the documentation.

Might be out of the scope of this PR as we should also discourage in-sources builds.



> The changes LGTM. However, there are currently no instructions to be able to use espp outside of the build directory (e.g. appending build dir to `PYTHONPATH`, setting -D CMAKE_INSTALL_PREFIX or --prefix, make install, etc).

Done rudimentarily.

PR #400, opened 2021-05-19
[FIX] build documentation

XzzX:


junghans:
Hmm, I would really like to solve this on the cmake level!

I thought the doc targets already depend on the python target!

For the python path we can use `cmake -E env PYTHON PATH=.... <Sphinx command>

XzzX:
> Hmm, I would really like to solve this on the cmake level!
Yes
> 
> I thought the doc targets already depend on the python target!
Python target?
> 
> For the python path we can use `cmake -E env PYTHON PATH=.... <Sphinx command>` as `COMMAND` to make that work.
I will try that.

junghans:
Haha all for this was there before the refactor https://github.com/espressopp/espressopp/blob/05ae583ddabb47e959f867fc0b7539c6af2bd599/CMakeLists.txt#L231-L236

junghans (changes requested):
Fix it on the CMake level instead.

XzzX:
> Haha all for this was there in v3.0.0 the refactor
> 
> https://github.com/espressopp/espressopp/blob/05ae583ddabb47e959f867fc0b7539c6af2bd599/CMakeLists.txt#L231-L236
> 
> Just restore the `${CMAKE_COMMAND} -E env "${TEST_ENV}"` and `DEPENDS` part ;-)

Yes, I also realized that. I wanted to get rid of all this hand-crafted stuff.

https://github.com/espressopp/espressopp/blob/b68d4c766c567a6c721d7df4a9f42328b02607fe/doc/sphinx/conf.py#L34-L39

would be the standard way. However, I do not know how to include the binary file there.

.github/workflows/validate.yml
```diff
@@ -77,10 +77,22 @@ jobs:
       - name: Install Python Dependencies
         run: pip3 install -r espressopp/requirements.txt
 
-      - name: build documentation
+      - name: build espressopp
         run: |
           cmake -S espressopp -B espressopp-build
-          cmake --build espressopp-build --target doxygen sphinx sphinx-pdf -j 2
```
junghans, line 83: Do we still need to split this now?

doc/CMakeLists.txt
```diff
@@ -19,20 +19,20 @@ find_program(SPHINX_EXECUTABLE
...
 
-if ( SPHINX_EXECUTABLE )
+if (SPHINX_EXECUTABLE)
     set(SPHINX_SOURCE ${CMAKE_CURRENT_SOURCE_DIR}/sphinx)
     set(SPHINX_BUILD ${CMAKE_CURRENT_BINARY_DIR}/sphinx)
 
     add_custom_target(sphinx
-            ${SPHINX_EXECUTABLE} -b html ${SPHINX_SOURCE} ${SPHINX_BUILD}/html
-            WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+            COMMAND ${CMAKE_COMMAND} -E env PYTHONPATH=${CMAKE_BINARY_DIR}:$ENV{PYTHONPATH} ${SPHINX_EXECUTABLE} -b html ${SPHINX_SOURCE} ${SPHINX_BUILD}/html
```
junghans, line 27: Is the module written to `${CMAKE_BINARY_DIR}`? And wouldn't it be better to use `TEST_ENV` here as well?

junghans: We might want to rename `TEST_ENV` to something more meaningful, like `ESP_PYTHON_ENV`.

junghans: Down the road, not in this PR.

.github/workflows/validate.yml
XzzX: No, but since the output is quite lengthy I find it easier to debug this way.

doc/CMakeLists.txt
XzzX: > Is the module written to `${CMAKE_BINARY_DIR}`? And wouldn't it be better to use `TEST_ENV` here as well?

We need `_espressopp.so` which is located in the build dir and the python files which are located in the source dir under `src` but renamend to `espressopp`. The former is done by using the build dir the later by using a symlink.

XzzX: > We might want to rename `TEST_ENV` to something more meaningful, like `ESP_PYTHON_ENV`.

Yes, `TEST_ENV` does not tell what it is doing.

junghans, line 27:
```suggestion
            COMMAND ${CMAKE_COMMAND} -E env PYTHONPATH=${TEST_ENV} ${SPHINX_EXECUTABLE} -b html ${SPHINX_SOURCE} ${SPHINX_BUILD}/html
```
And I will replace `TEST_ENV` later everywhere.

XzzX: ok

junghans (approved):

PR #401, opened 2021-05-19
installation via pip

XzzX:
alternative installation method

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -59,7 +59,7 @@ jobs:
         run: cmake --build build --verbose --target all -j 2
 
       - name: Install
-        run: cmake --install build
+        run: pip3 install -v build/.
 
       - name: Test
         working-directory: build
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,3 +131,5 @@ add_custom_command(TARGET symlink COMMAND ${CMAKE_COMMAND} -E create_symlink
         ${CMAKE_SOURCE_DIR}/src espressopp)
 
 add_subdirectory(doc)
+
+configure_file(${CMAKE_SOURCE_DIR}/setup.py ${CMAKE_BINARY_DIR}/. COPYONLY)
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,11 @@
+from setuptools import setup, Extension
+
+print(setuptools.find_packages(where="."))
+
+setup(
+    name='espressopp',
+    version='3.0.0',
+    packages=setuptools.find_packages(where="."),
+    package_dir={'': '.', 'espressopp': 'espressopp'},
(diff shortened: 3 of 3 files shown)


.github/workflows/validate.yml
```diff
@@ -59,7 +59,7 @@ jobs:
         run: cmake --build build --verbose --target all -j 2
 
       - name: Install
-        run: cmake --install build
+        run: pip3 install -v build/.
```
junghans, line 62: I would also like to test both install mechanism somehow.

XzzX: I do not like duplication. One has to maintain two things and they can diverge. I prefer to stick to one way of installing espressopp, either via `make install` or `pip`.

setup.py
```diff
@@ -0,0 +1,11 @@
+from setuptools import setup, Extension
+
+print(setuptools.find_packages(where="."))
+
+setup(
+    name='espressopp',
+    version='3.0.0',
```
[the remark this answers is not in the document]
XzzX: Yes, same for project name, etc. And we should supply more meta information. This was just a quick draft. :)

CMakeLists.txt
```diff
@@ -131,3 +131,5 @@ add_custom_command(TARGET symlink COMMAND ${CMAKE_COMMAND} -E create_symlink
         ${CMAKE_SOURCE_DIR}/src espressopp)
 
 add_subdirectory(doc)
+
+configure_file(${CMAKE_SOURCE_DIR}/setup.py ${CMAKE_BINARY_DIR}/. COPYONLY)
```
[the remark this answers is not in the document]
XzzX: We can rename the not processed file.

.github/workflows/validate.yml
junghans: Spack and the rpm packages on openSUSE and fedora (https://github.com/espressopp/fedora-copr) use cmake to install, so I would really like to keep that.

XzzX: I am not familiar in building spack or rpm packages. If they cannot use `pip` then let's stick to our hand-crafted `make install`.

junghans: Everything is possible in spack and rpm, using standard cmake just makes it much easier as spack has a cmake class and rpm has cmake macros.

Plus our `make install` is pretty standard already. E.g. it supports `DESTDIR`.

XzzX: Ok.

XzzX: Do we want to support both installation methods? Then we have to explain what the difference between `make install` and `pip install` is and keep it up-to-date.

junghans:
@XzzX now obsolete?

XzzX:
obsolete

Issue #369, opened 2021-05-04
Male sure examples actually work


junghans:
Some of them haven't been touched for years.

We might at least want to byte compile them.

XzzX:
What about running them on a weekly basis?

junghans:
Sounds good.

They aren't integrated in CMake though.

XzzX:
True, we could add them as tests with a specific label, if you want to run them via cmake. I will have a look at it.

PR #404, opened 2021-05-20
introduce python auto formatting

XzzX:
Any suggestions for a PEP8 config file?


format.sh
```diff
@@ -1,3 +1,4 @@
 #! /bin/bash
 
 git ls-files -- src*.cpp src*.hpp testsuite*.cpp testsuite*.hpp | xargs clang-format -i -style=file
+git ls-files -- src*.py testsuite*.py | xargs autopep8 --in-place --aggressive --aggressive
```
jnvance, line 4: Maybe also include `examples*.py`, or should it be done after #369?

jkrajniak:
> Any suggestions for a PEP8 config file?

yes, let's not stick to 80 characters, I think it is still in the standard but it's really annoying and old - 100, 120 seems reasonable, IMO.

XzzX:
> > Any suggestions for a PEP8 config file?
> 
> yes, let's not stick to 80 characters, I think it is still in the standard but it's really annoying and old - 100, 120 seems reasonable, IMO.

Good point, should be the same as for C++.

format.sh
XzzX: I think both are independent. I will include examples.

XzzX:
We have circular imports...

PR #409, opened 2021-09-17
fix build on openSUSE

### LINKED ISSUE #408: WEEKLY CI jobs are broken
OpenSUSE has problems with the current integration of GROMACS:

https://github.com/espressopp/espressopp/runs/3510954362?check_suite_focus=true

junghans:
Fix #408

### DIFF
--- a/src/analysis/OrderParameter.hpp
+++ b/src/analysis/OrderParameter.hpp
@@ -46,6 +46,7 @@
 // #include <boost/math/special_functions/spherical_harmonic.hpp>
 
 #include <algorithm>
+#include <unordered_map>
 
 using namespace std;
 using namespace boost;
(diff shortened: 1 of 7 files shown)


junghans:
@XzzX we should revert 792d3ada97e42f15480374a4c070c6b4d4b6937e after this PR and figure out a better way to do the comparison, but my guess is that the internals of the xtc format changed and hence the files aren't binary the same anymore.

junghans:
> @XzzX we should revert [792d3ad](https://github.com/espressopp/espressopp/commit/792d3ada97e42f15480374a4c070c6b4d4b6937e) after this PR and figure out a better way to do the comparison, but my guess is that the internals of the xtc format changed and hence the files aren't binary the same anymore.

I guess the best would be to actually read both files and compare, but for that we need to implement `readxtc()` first.

junghans:
Manually trigger weekly build was successful: https://github.com/espressopp/espressopp/actions/runs/1247538449

XzzX (approved):

PR #412, opened 2021-09-24
Add secomp options

junghans:
Related to https://github.com/espressopp/buildenv/pull/57


XzzX (approved):

junghans:
@XzzX I have no good idea were the segfault are coming from.

PR #413, opened 2021-09-27
Reverse mapping

jkrajniak:
This PR adds code that was used to perform reverse mapping of different systems presented in:
 - Generic Adaptive Resolution Method for Reverse Mapping of Polymers from Coarse-Grained to Atomistic Descriptions (https://pubs.acs.org/doi/abs/10.1021/acs.jctc.6b00595)
 -  Reverse mapping method for complex polymer systems (https://onlinelibrary.wiley.com/doi/abs/10.1002/jcc.25129)

Although the method uses AdResS, in fact it works a bit different from the implementation that is already in E++. Basically, all particls (both AT and CG) are present in the simulation box. One additional parameter to the Particle was added: vp (virtual particle). The relationship between CG and AT particles is kept in FixedVSList.

### DIFF
--- a/src/FixedVSList.cpp
+++ b/src/FixedVSList.cpp
@@ -1,0 +1,76 @@
+/*
+  Copyright (C) 2014-2017
+      Jakub Krajniak (jkrajniak at gmail.com)
+
+  This file is part of ESPResSo++.
+
+  ESPResSo++ is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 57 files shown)


jnvance:
Hi Jakub, do you plan to add unit or integration tests for the newly added classes?

jkrajniak:
> Hi Jakub, do you plan to add unit or integration tests for the newly added classes?

Good point! I will add some unit tests

PR #422, opened 2022-02-17
pull version from cmake

### LINKED ISSUE #421: Generate version in version.hpp from CMake

junghans:
Fix #421


cmake/gitversion.cmake
```diff
@@ -21,6 +21,9 @@ endif (GIT_FOUND AND IS_DIRECTORY ${TOP_SOURCE_DIR}/.git)
 set (GIT_HEADER "gitversion.hpp")
 set (NEW_GIT_HEADER "new_gitversion.hpp")
 file(WRITE ${NEW_GIT_HEADER} "static const std::string gitversion = \"${ESPP_GIT_ID}\";\n")
+file(APPEND ${NEW_GIT_HEADER} "#define MAJORVERSION ${PROJECT_VERSION_MAJOR}\n")
+file(APPEND ${NEW_GIT_HEADER} "#define MINORVERSION ${PROJECT_VERSION_MINOR}\n")
+file(APPEND ${NEW_GIT_HEADER} "#define PATCHLEVEL ${PROJECT_VERSION_PATCH}\n")
```
XzzX, line 26: I generally do not like defines very much. Can we use `constexpr int`?

junghans: Currently these aren't used anywhere, so we could do that!

XzzX (approved):

PR #424, opened 2022-03-22
Replicate in parallel

jnvance:
Added `espressopp.tools.ReplicateParallel` which performs the replicate operation in parallel on all workers (instead of doing it on the controller) and creates the replicated particles, bonds and angles only when they are being added to the respective classes. They are also instantiated by batches of the same size as the original seed so memory footprint is reduced.


xzhh:
Ok

PR #446, opened 2022-12-20
cmake: only enable C for old cmake versions

junghans:


XzzX (commented):
Do we have a CI job that tests CMake 3.26+?

junghans:
Fedora rawhide has CMake-3.28.3 now.

junghans:
Still failing?!

I thought https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8015 is in there now.

junghans:
@XzzX ready for review.

XzzX (approved):
Might break for older CMake versions. Why did we have our own FindHDF5.cmake in the first place? If I remember correctly, I added it because HDF5 targets were added pretty late.

junghans:
> Might break for older CMake versions. Why did we have our own FindHDF5.cmake in the first place? If I remember correctly, I added it because HDF5 targets were added pretty late.

We had it to get the `HDF5` target.

junghans:
We could increase the cmake version required to the first version that had the `hdf5` target. (Looking at https://github.com/Kitware/CMake/commit/a8e0a6b3e4407d23c3a289ac5b0a782e745bccc3, cmake 3.19 had it.)

Or update the bundled module.

PR #491, opened 2023-06-15
Adapting for recent changes in HPX

hkaiser:
@junghans: As requested, this applies changes caused by recent API modifications in HPX. There are no big surprises, except for the fact, that for whatever reason the HPX error names were not exposed publicly anymore (which is a bug). I have created a PR on the HPX side (see: https://github.com/STEllAR-GROUP/hpx/pull/6279) that reintroduces this functionality. Please note that this PR here relies on the newly added API function, which will be available with the upcoming HPX V1.9.1 release only.

PS: I was not sure whether you preferred having these changes applied conditionally. For now this PR applies the changes unconditionally, that however should give you a hint what needs changing.

### DIFF
--- a/src/hpx4espp/HPXRuntime.cpp
+++ b/src/hpx4espp/HPXRuntime.cpp
@@ -34,8 +34,6 @@
 #include <hpx/runtime/agas/addressing_service.hpp>
 #endif
 
-#include <hpx/runtime/parcelset/parcelhandler.hpp>
-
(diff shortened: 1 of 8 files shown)


junghans:
@hkaiser thanks, we can wait for v1.9.1 that won't be a problem.

junghans (approved):

junghans (approved):

Issue #514, opened 2026-08-26
Suggestions from the JOSS review


junghans:
Items from the JOSS Review (see https://github.com/openjournals/joss-reviews/issues/11072#issuecomment-5426333494):
- [x] the user guide is available as both GitHub Pages and a PDF file, but the latter contains more information than the former, in particular the GitHub Pages are missing installation instructions and the Python API docstrings (#525)
- [x] installation instructions
   * [x] matplotlib instructions point to the sourceforge package, but it hasn't received updates since 2017; I would recommend pointing users to Linux packages or PyPi (#520)
   * [x] the instructions don't mention that Python dependencies can be installed via `pip install -r requirements.txt` or similar (#521)
   * [x] the instructions ask to run `export PYTHONPATH=$PWD/src:${PYTHONPATH}`, but:
      1. [x] if `PYTHONPATH` is empty, there will be a trailing colon in the environment variable that will be interpreted as the current directory, and this can lead to really cryptic error messages; it is more robust to make the colon conditional on an expansion like `export PYTHONPATH="${PWD}/src${PYTHONPATH:+:$PYTHONPATH}"` (#517)
      2. [x] this environment variable didn't work for me: `python -c "import espressopp"` exited with `ModuleNotFoundError: No module named 'espressopp'`; instead, I had to execute `cmake --install builddir --prefix $VIRTUAL_ENV` to install the files to my Python environment (#523)
   * [x] when importing `espressopp` for the first time, .pyc files are created and several `SyntaxWarning: invalid escape sequence` messages are displayed; consider prefixing docstrings with `r"""` to make them raw string (which allows `\g`, `\s`, `\*` and prevents `\t` from silently becoming tabulation characters), or escaping every backslash with `\\` in docstrings (#522, already fixed in 01cd7fcd5d0467e6a1e2389a7e712b6cf2e15df0)
* [x] manuscript formatting
   * [x] missing white space between text and some references, e.g. "systems(Praprotnik" and "codes(Fritsch" (#519)
   * [x] there are Spanish question marks in the volume and issue information, or at least, it's how these Unicode characters are - displayed on my end - that will fix itself later
* [x] in `CONTRIBUTING.md` (#515):
   - [x] missing e-mail address in `sensitive bugs must be sent by email to <>.`
   - [x] consider replacing `http` by `https` in `http://espressopp.github.io`
[...]

PR #525, opened 2026-08-27
doc: fix html gen

junghans:
Part of #514


junghans:
`PYTHONPATH` was not set correctly for the `sphinx` target hence documentation was not fully parsed, the `sphinx-pdf` target was fine however.

